PR #3, opened 2014-08-16
[POC] Behat

dantleech:
This PR executes documentation steps from `installation.rst` with [Behat](http://behat.org).

It makes use of the [shinx-behat](https://github.com/dantleech/sphinx-behat) Sphinx extension which generates behat `.feature` files from annotations within the documentation.

Behat will install SuluCMF in a `_workspace` directory and execute the console, based on the installation instructions, verifying that every command executed was successful.

See: https://github.com/sulu-cmf/sulu-docs/blob/behat/getting-started/installation.rst as an example of the annotation syntax.

The generated fixture file looks like: https://gist.github.com/dantleech/69eec3aaffa021c6c869

### DIFF
--- a/_behat/FeatureContext.php
+++ b/_behat/FeatureContext.php
@@ -1,0 +1,134 @@
+<?php
+
+require_once(__DIR__.'/../vendor/autoload.php');
+
+use Behat\Behat\Context\ClosuredContextInterface,
+    Behat\Behat\Context\TranslatedContextInterface,
+    Behat\Behat\Context\BehatContext,
+    Behat\Behat\Exception\PendingException;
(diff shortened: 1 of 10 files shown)


_behat/FeatureContext.php
```diff
@@ -0,0 +1,134 @@
...
+// Require 3rd-party libraries here:
+//
+//   require_once 'PHPUnit/Autoload.php';
+//   require_once 'PHPUnit/Framework/Assert/Functions.php';
+//
+
+/**
+ * Features context.
+ */
+class FeatureContext extends BehatContext
```
dantleech: No idea why this file is here

dantleech:
Closing this for now.

PR #5, opened 2014-11-28
Developer book

dantleech:
I think we should add a developer guide here. This would be a separate book, so there would be the User section (which I believe is what this currently is) and the Developer section.

We need this because developer documentation should be a first-class citizen and the quality of such a document should be important otherwise it will not be used.

For example I want to write a guide on how to write behat tests, standardize UI workflows (e.g. there should always be a success / error notification, delete buttons should always be here, etc.).

Doing this in markdown in the existing `/docs` repository which is of variable quality does not really appeal to me and browsing such documents on GITHub isn't the best experience - RST provides a much richer medium for technical documentation and readthedocs provides a great reading experience.

### DIFF
--- a/developer/behat/getting-started.rst
+++ b/developer/behat/getting-started.rst
@@ -1,0 +1,114 @@
+Getting started
+===============
+
+Choosing Driver
+---------------
+
+Selenium2 (with a browser)
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is recommended to run tests locally with the `Selenium`_ driver. Using the
+Selenium driver on a real browser allows you to see the tests running (as
+opposed to using a headless browser as detailed in the following sections).
+
+You can `download Selenium server from here`_, or install it as follows:
+
+.. code-block:: bash
+
+    $ mkdir ~/jars
+    $ cd jars
+    $ wget http://selenium-release.storage.googleapis.com/2.44/selenium-server-standalone-2.44.0.jar
+
+You can then run it:
+
+.. code-block:: bash
+
+    $ java -jar selenium-server-standalone-2.44.0.jar -browserSessionReuse -singleWindow
+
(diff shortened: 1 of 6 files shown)


developer/behat/getting-started.rst
```diff
@@ -0,0 +1,118 @@
...
+    You will need to install the Java Runtime Environment. On a debian based
+    system you could do ``apt-get install default-jre``
+
+Selenium2 (headless, with PhantomJs)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note::
+
+    It is not currently recommended to run tests on PhantomJS - the results
+    maybe unpredictable.
```
danrot: shouldn't it be `may be`?

developer/behat/getting-started.rst
```diff
@@ -0,0 +1,118 @@
...
+            # ...
+            Behat\MinkExtension:
+                sessions:
+                    default:
+                        selenium2:
+                            wd_host: "http://localhost:8643/wd/hub"
+
+.. note::
+
+    Tests run with phantom JS may be unstable.
```
danrot: duplicate note of above?

developer/behat/getting-started.rst
```diff
@@ -0,0 +1,118 @@
...
+Running the tests
+-----------------
+
+You can run all the tests as follows:
+
+.. code-block:: bash
+
+    $ ./vendor/behat/behat/bin/behat -p<profile>
+
+Where ``<profile`` is one of:
```
danrot: `<profile>`

developer/behat/getting-started.rst
dantleech: Yes

dantleech:
Updated.

PR #9, opened 2015-02-02
Added command for creating role to getting started guide

danrot:


book/getting-started/setup.rst
```diff
@@ -146,9 +146,17 @@ configuration files) and loads the fixtures:
 
 Create a new user
 -----------------
-In order to login into Sulu you need to create an user. Therefore you just have
-to enter the following command on the command line, which will guide you
-through the creation in an interactive manner:
+In order to login into Sulu you need to create an user. Before you can do that
```
dantleech: Always have a blank line after a heading

dantleech: create *a* user

book/getting-started/setup.rst
```diff
@@ -146,9 +146,17 @@ configuration files) and loads the fixtures:
...
+you have to create the administrator role. You can easily add this role with
+the following command:
+
+.. code-block:: bash
+
+    app/console sulu:security:role:create
```
dantleech: use

````
.. code-block: bash

    $ app/console sulu:security:role:create
````

dantleech: And I am not sure if we should always prefix `php` before `app/console` i.e. `$ php app/console`. But whatever we do we should make that the standard.

PR #10, opened 2015-02-09
Glossary

dantleech:
This PR is a work in progress which aims to set a standard set of terms which we will use in Sulu, and remove any ambiguous terminology in the future.


dantleech:
Merged this into the `document-manager` branch.

PR #23, opened 2015-03-13
Add scripts, iframes, height, maxHeight, enterMode Options to documentation

alexander-schranz:


reference/content-types/text_editor.rst
```diff
@@ -22,6 +22,21 @@ Parameters
...
+      - boolean
+      - Allow iframe tags in source code
+    * - height
+      - integer
+      - Set the text editor height
+    * - maxHeight
+      - integer
+      - Set the max height for autogrow '0' is  infinitely
+    * - enterMode
+      - string P/BR/DIV
```
dantleech: I think this should simply be `enum`. The values should be given in the description.

reference/content-types/text_editor.rst
```diff
@@ -22,6 +22,21 @@ Parameters
...
+      - Set what happens on enter allowed P, BR or DIV
```
dantleech: you must put "code words" in double back-ticks. This both helps distinguish them from normal text (for humans) and also for the parser for when we add spell checking.

````
Set what happens, one of: ``p``, ``br`` or ``div``
````

wachterjohannes:
@alexander-schranz ping

alexander-schranz:
closed enter_mode ... are ckeditor specific parameters and example for ckeditor exist

PR #26, opened 2015-04-21
Added new admin sentences

wachterjohannes:


wachterjohannes:
@dantleech @danrot ping

danrot:
@dantleech Can you have a look if this makes sense?

developer/behat/context-reference.rst
```diff
@@ -48,6 +48,15 @@ The following is not a fully comprehensive list:
...
 - ``I wait for the ajax request``: Wait until all underlying AJAX requests
   have finished.
+- ``I click the close icon``: Click the close icon (``.fa-times``)
+- ``I click the close icon in container ":selector"``: Click the close 
+  icon (``selector + ' .fa-times'``)
+- ``I wait for the column navigation column :index``: Wait for the column 
+  navigation column with given id.
+- ``I expect a data-navigation to appear``: Wait and assert that a 
+  data-navigation will appear.
+- ``I expect a overlay to appear``: Wait and assert that a overlay
```
dantleech: `s/a/an`

dantleech:
In general we shouldn't reference CSS classes in Behat, the sentances should be understandable by non-coders. But this is something we need to improve in general.

developer/behat/context-reference.rst
```diff
@@ -48,6 +48,15 @@ The following is not a fully comprehensive list:
 - ``There should be errors``: Assert that there are errors on the page.
 - ``I wait for the ajax request``: Wait until all underlying AJAX requests
   have finished.
+- ``I click the close icon``: Click the close icon (``.fa-times``)
+- ``I click the close icon in container ":selector"``: Click the close 
+  icon (``selector + ' .fa-times'``)
+- ``I wait for the column navigation column :index``: Wait for the column 
+  navigation column with given id.
+- ``I expect a data-navigation to appear``: Wait and assert that a
```
dantleech: This should be both `expect` and `wait for` in the context.

dantleech:
tbh, I am not so sure that this it is good to document the sentances at all.

wachterjohannes:
@dantleech fixed and fixed typos in `adminContext` https://github.com/sulu-io/sulu/pull/1236

danrot:
@dantleech You already started it, why should we stop doing it?

wachterjohannes:
@dantleech where should i then know which sentences are possible?

dantleech:
@danrot the syntax is already bad and there are multiple variations of each "sentance", in short its just quite a lot of maintenance.

@wachterjohannes from the code .. but if you use the current docs you already do not know which sentances are possible. The best thing would be a command to list all of the sentances per context. Might make a good PR for behat.

But I don't care either way.

wachterjohannes:
hmm a command would be the best i think ... i dont think that anyone knows that he has to look in the context for the possible sentences! so until we have no other solution we should at least try to keep it actual!

wachterjohannes:
@dantleech everything would be fixed here!

wachterjohannes:
@dantleech ping

PR #33, opened 2015-05-22
Updated config parameters

eiannone:


dantleech:
Good catch :) Thanks.

PR #34, opened 2015-05-22
Added installation setup for windows OS

eiannone:


danrot:
@sabinebaer would be awesome if you could have a look at this, and tell us if this works correctly :-)

danrot:
@eiannone What I can see already from the build is that `dos` doesn't exist as a code block type. Maybe you should use bash instead? Or is there any other type suited better? /cc @dantleech

book/getting-started/setup.rst
```diff
@@ -149,6 +149,21 @@ Or these commands for Mac OSX:
     sudo chmod +a "$APACHEUSER allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs uploads/media web/uploads/media app/data
     sudo chmod +a "`whoami` allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs uploads/media web/uploads/media app/data
 
+Or these commands for Windows (with IIS web server):
+
+.. code-block:: dos
```
dantleech: Maybe try `dosbatch` instead? http://pygments.org/docs/lexers/#pygments.lexers.shell.BatchLexer

eiannone:
Changed code block type to `batch`

danrot:
Yeah, squashing would be nice.
After @sabinebaer has tested it, it can be merged.

Thank you for your support :smile:

sabinebaer:
The commands can be executed with Windows CMD but not with Windows PowerShell (not sure if anybody out there is using PowerShell or if it is obvious for them that this only works with CMD).

The behaviour is as expected, folders are recreated, created and full access is granted to IUSR.

However, I did not test if these are the correct settings for IIS (as I do not have IIS installed at the moment).

danrot:
Which commands don't work in the powershell? Just assuming that something is obvious doesn't sound too good, I would say we should mention that it is only working with the Windows CMD. /cc @eiannone

alexander-schranz:
@danrot @sabinebaer @eiannone I think the only thing which will not work in powershell is the  '&&' operator if you put them in seperate lines it will work in powershell and cmd.

sabinebaer:
@alexander-schranz I tried this but /Q is also not working (Remove-Item : A positional parameter cannot be found that accepts argument '/Q'.)

sabinebaer:
@danrot Only the md command is working, rd and icacls throw errors. I think mentioning that this only works for Windows CMD is enough.

alexander-schranz:
@sabinebaer you are right but we can run them under cmd with `cmd /c "rd /S /Q app\cache && md app\cache"` which works in cmd and powershell

eiannone:
Do you think it would be more appropriate to use powershell instead of standard cmd?
I can write a powershell version of the script if necessary.

danrot:
Actually I haven't been using windows for some time, so I have no idea what is more appropriate :-) /cc @sabinebaer @alexander-schranz 

@alexander-schranz BTW, haven't you implemented once that the `uploads/media` and `web/uploads/media` directories are created automatically? Then we should remove these lines from the list for all three operating systems.

alexander-schranz:
My experience is that powershell is used more often then cmd, because the microsoft official tutorials mostly based on powershell not on cmd.

@danrot thats right they are created by the mediabundle init command.
only the permissions needed to be set correctly for this folders

eiannone:
So I have converted the script to powershell, and removed the redundant lines for already created folders (also on Linux script).

danrot:
Awesome, thank you :grinning: so if @sabinebaer or @alexander-schranz can test it again and it works it is fine to merge for me.

sabinebaer:
@danrot @eiannone The script can be executed with powershell and works as expected.

danrot:
@eiannone Thank you :grinning:

PR #42, opened 2015-06-24
Document manager query builder

dantleech:
This will be the branch with the query builder and mapping documentation.


alexander-schranz:
@danrot the most of this PR exists currently in master branch the only one which is not in master is the [mapping.rst](https://github.com/sulu/sulu-docs/pull/42/files#diff-dbf8190646de5767a8e46b5ce51d2e30) not sure if we still need this?

danrot:
Yeah, and I think it is also outdated anyway... So I am closing it.

PR #57, opened 2015-07-21
Installation instructions: Clone url with HTTPS

whelmich:
Use the HTTPS clone url as recommended by GitHub, described here: https://help.github.com/articles/which-remote-url-should-i-use/


danrot:
Makes sense, thank you!

PR #64, opened 2015-08-17
Added tags_parameter option to SmartContent and tag twig functions

wachterjohannes:
Main PR https://github.com/sulu-io/sulu/pull/1538

### DIFF
--- a/reference/twig-extensions/functions/sulu_tags.rst
+++ b/reference/twig-extensions/functions/sulu_tags.rst
@@ -1,7 +1,7 @@
 ``sulu_tags``
 =============
 
-Returns all tags.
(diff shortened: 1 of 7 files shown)


reference/twig-extensions/functions/sulu_tags.rst
```diff
@@ -0,0 +1,15 @@
+``sulu_tags``
+=============
+
+Returns all tags.
```
danrot: All tags in the system? All tags in the current request? Is not completely clear to me (maybe if you look at the example, but I would still specify that further here).

danrot:
else ACK

PR #97, opened 2015-09-10
documented headline parameter

marcelmoosbrugger:


reference/content-types/text_line.rst
```diff
@@ -20,4 +28,7 @@ Example
         <meta>
             <title lang="en">Title</title>
         </meta>
+        <params>
+            <param name="headline" value="true"/>
```
wachterjohannes: why you dont use a tag for this feature?

danrot: @marcelmoosbrugger Curious about that one two, that's what we have actually agreed on :wink:

reference/content-types/text_line.rst
```diff
@@ -9,7 +9,15 @@ Shows a simple text line, the inserted content will be saved as simple string.
...
-No parameters available
+.. list-table::
+:header-rows: 1
+
+    * - Parameter
+      - Type
+      - Description
+    * - headline
+      - boolean
+      - Makes the text line identifiable as the title of the content
```
danrot: @marcelmoosbrugger If this has just a visual effect this description is wrong.

reference/content-types/text_line.rst
marcelmoosbrugger: tag's seem like something which can be put on different content-types. The headline parameter only has effects on the text line. Its so to say a parameter of the text line. Because in other content types, like the text-editor or the media-selection, parameters were used in the same sense, @chirimoya told me to implement it as a parameter

joemoe:
@danrot and @wachterjohannes is this ok now?

wachterjohannes:
for me its ok!

PR #103, opened 2015-09-17
Fixed wrong file path in setup instructions

tschreiner:
The following line in the setup instructions is wrong:
```cp app/Resources/webspaces/sulu.io.xml.dist app/Resources/webspaces/sulu.io.xmlre```

I changed the extension to ```xml```


tschreiner:
I selected the wrong target branch. sorry

PR #118, opened 2015-10-06
added documentation for media security

danrot:


bundles/media.rst
```diff
@@ -0,0 +1,22 @@
...
+apply to the assets contained in the collection.
+
+So the `view` permission is required to see the collection itself in the
+navigation and its assets in any available view in the system.
+
+The `add` permission is needed to upload new assets and add new subcollections
+to a collection. With the `edit` permission the user is allowed to edit the
+title and other attributes of the collections and its containing assets, and
+finally the `delete` permission is required to delete assets and entire
+collections.
```
wachterjohannes: and remove images from the collection ??

danrot: That's what I mean by assets, wrote it like this because it is not only about images (could also be documents, ...)

wachterjohannes: your right (: i should read it more carefully (:

PR #120, opened 2015-10-11
Add snippetType option to snippet content type

lennerd:
This PR adds the `snippetType` option to the `snippet` content type reference.


danrot:
Thank you :smiley:

PR #140, opened 2015-12-01
added guide for releasing

danrot:


danrot:
@sulu-io/core-developer What should we do about this? I would say we also try to update this list while doing the 1.2 release, and merge it afterwards.

webmozart:
@danrot :+1:

PR #146, opened 2016-01-07
Added webspace-settings to doc

wachterjohannes:
main https://github.com/sulu-io/sulu/pull/1912


cookbook/manage-analytics.rst
```diff
@@ -0,0 +1,32 @@
+How to manage analytics?
+========================
+
+Sulu gives the content-manger an easy way to manage analytic-codes and appends
+them automatically to the website output without any changes in the
+twig-template. You can find the list of analytics under the webspace-settings.
+
+The Analytic consists of:
```
danrot: `The analytics consist of`

cookbook/manage-analytics.rst
```diff
@@ -0,0 +1,32 @@
...
+      - The code or key of the analytic.
+
+Sulu can handle different types of analytic-systems like google or piwik.
+This codes will be automatically added with the given key and site-id (for
+piwik). To add other systems simply choose type custom and copy and paste
+the code into the textarea.
+
+.. warning::
+
+    Be aware of that custom analytics will not be evaluated and appended without
```
danrot: `Be aware that` - and shouldn't the code at least be validated on its own?

wachterjohannes: but what would you evaluate? correct javascript? you could write for example `$('html`).html('')` this would clear the whole content. but it is correct javascript .... therefor its a "bad" feature and should at least be warned here.

PR #151, opened 2016-01-11
[DX] Add concrete exemple to use fixtures

cbastienbaron:
use Sulu\Component\DocumentManager\DocumentManager;


wachterjohannes:
@danrot what do you think? its ok for me!

danrot:
@wachterjohannes If it works it is ok for me. Can you maybe test it?

PR #171, opened 2016-02-24
Update set permission for installation

alexander-schranz:
Inspired by Symfony: http://symfony.com/doc/2.8/book/installation.html#book-installation-permissions

### DIFF
--- a/book/getting-started/setup.rst
+++ b/book/getting-started/setup.rst
@@ -161,8 +161,9 @@ Use the following commands for Linux:
     rm -rf app/cache/*
     rm -rf app/logs/*
     mkdir app/data
-    sudo setfacl -R -m u:www-data:rwx -m u:`whoami`:rwx app/cache app/logs uploads/media web/uploads/media app/data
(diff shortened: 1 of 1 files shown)


danrot:
The mac osx variant works, will test the linux command later before merging.

danrot:
@alexander-schranz If you are already touching this stuff, can you also integrate the suggested change in https://github.com/sulu-io/sulu/pull/2001#issuecomment-183361824 ?

alexander-schranz:
would set the `/uploads/media` also set to `/uploads` so a developer can create their his own folders to store files what you think @danrot

danrot:
That's what we have agreed on in the other discussion, no?

alexander-schranz:
notsure, but if this is ok branch is rebased an ready to merge.

patkar:
Btw. #166 already changed `/uploads/media` to `/media` too on the develop ;)

danrot:
True, but then it looks like this branch was created before that.

PR #172, opened 2016-02-25
Moved parameter documentation to reference

webmozart:
This PR moves the description of the parameters to the reference section in order to shorten the "Getting Started" guide.

In a follow-up PR, I'd like to add a section for each parameter with an example and more specific documentation.

Open questions:

- [x] Why do we have "currently unused" parameters? Can we remove those?
- [x] Why are the two files separate? Can we merge parameters.yml and phpcr.yml?

### DIFF
--- a/book/getting-started/installation.rst
+++ b/book/getting-started/installation.rst
@@ -26,69 +26,15 @@ checkout the latest version of Sulu:
 Install dependencies
 --------------------
 
-We use `composer`_ to install the correct versions of
-Sulu's dependencies:
+Use `Composer`_ to install Sulu's dependencies:
 
 .. code-block:: bash
 
     composer install
 
-At the end of the installation you will be asked for some parameters. The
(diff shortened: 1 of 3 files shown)


danrot:
Looks good to me. Regarding your questions:

* We have these because they are in the symfony standard edition. So yes, we can probably remove those.
* The symfony-cmf also split these two files, if I remember correctly. But for me it would also be fine to merge them, as long as nobody knows about any concrete advantage.

webmozart:
Answer 1: The parameters can not be removed. The Sulu Standard edition should resemble Symfony Standard closely. Hence SwiftMailer should remove part of Sulu Standard and these parameters need to remain. The text "not currently used" can be removed.

Answer 2: The files can be merged. See sulu-io/sulu#2068

wachterjohannes:
looks good to me

PR #184, opened 2016-03-04
How to resize within a container

ollietb:
There were some missing options from the docs about how to constrain the image size within a box without scaling


alexander-schranz:
:+1:

chirimoya:
@ollietb thx

PR #202, opened 2016-04-07
Adapted permission settings to be more general

danrot:
This PR updates the commands for setting permissions, to remove some coupling from the media bundle here.

@sabinebaer @alexander-schranz: Could one of you also test the windows commands? Especially if they are still working after executing `app/console sulu:media:format:cache:clear`.


danrot:
Closed in favour of #171

PR #203, opened 2016-04-08
Add some background on domain source

tobias47n9e:
The current text does not explain how the domain is chosen.
Adding a sentence to clarify how it is constructed.

### DIFF
--- a/reference/twig-extensions/functions/sulu_content_path.rst
+++ b/reference/twig-extensions/functions/sulu_content_path.rst
@@ -1,7 +1,7 @@
 ``sulu_content_path``
 =====================
 
-Returns the absolute URL for the content at the given path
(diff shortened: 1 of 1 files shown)


reference/twig-extensions/functions/sulu_content_path.rst
```diff
@@ -1,7 +1,7 @@
 ``sulu_content_path``
 =====================
 
-Returns the absolute URL for the content at the given path
+Returns the absolute URL for the content at the given path. The domain is taken from ``app/Resources/webspaces/*.io.xml`` and your current environment. In case you have multiple URLs in one environment, you can prioritize one by giving it ``<url main="true">``.
```
danrot: You should adhere to our documentation standards here, which say that one line has a max length of 80 character.

PR #205, opened 2016-04-12
Fixed navigation twig extensions

danrot:
One of the title of the navigation twig functions was wrong, and the descriptions for the non-root methods were not correct.


reference/twig-extensions/functions/sulu_navigation_flat.rst
```diff
@@ -1,10 +1,11 @@
-``sulu_navigation_root_flat``
+``sulu_navigation_flat``
 =============================
```
wachterjohannes: shorten this line

PR #207, opened 2016-04-12
Show how to get link from target UUID

tobias47n9e:


wachterjohannes:
@dantleech can you check the spelling technically it is ok

dantleech:
thanks @tobias47n9e !

PR #210, opened 2016-04-13
Add paragraph about string localization

tobias47n9e:


book/creating-a-basic-website/adding-localizations.rst
```diff
@@ -56,3 +56,39 @@ the content. In case there is a ghost page - which means that the page is not
 translated into the current localization - this tree will be used to determine
 the "closest" language available.
 
+Localizing Strings
```
wachterjohannes: this shouldn't be in this doc-file because this file handles backend locales not website translations. Also i am not should if we should describe symfony features in our docs /cc @danrot @webmozart @chirimoya

danrot: That's what I also said... I think explaining pure symfony features in our docs doesn't make too much sense.

danrot:
@tobias47n9e Can we maybe only link the symfony documentation? But then we would also have to split that part into admin and website.

danrot:
@alexander-schranz You are right, documenting features already explained in Symfony would just be duplicating stuff, and if they change something we would probably forget to update it in our docs...

PR #211, opened 2016-04-15
Added php extension requirements

wachterjohannes:
| Q | A
| --- | ---
| Fixed tickets | none
| License | MIT

#### What's in this PR?

This PR adds the requirements for php-extensions.

#### Why?

@mob8607 had problems with installing sulu on his private server. after installing this two extension it works.

### DIFF
--- a/book/getting-started/system-requirements.rst
+++ b/book/getting-started/system-requirements.rst
@@ -7,7 +7,7 @@ For successfully running Sulu you'll have to fulfill the following system requir
 
 * Mac OSX, Linux or Windows
 * Webserver (`apache <http://httpd.apache.org/>`_ or `nginx <http://nginx.org/>`_ with enabled URL rewriting)
-* `PHP <http://php.net/>`_ 5.5 or higher
+* `PHP <http://php.net/>`_ 5.5 or higher with php-intl and one of the image-libraries php-gd or imagick
 * Database (`MySQL <http://www.mysql.com/>`_, `PostgreSQL <http://www.postgresql.org/>`_ or any other database supported by `doctrine <http://www.doctrine-project.org/>`_)
 * `composer <https://getcomposer.org/>`_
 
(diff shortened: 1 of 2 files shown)


wachterjohannes:
also he thinks that we should add a hint that the default template uses assetic and if the demo should be run in production `assetic:dump` has to be called. @danrot where should i add this hint?

book/getting-started/system-requirements.rst
```diff
@@ -7,7 +7,7 @@ For successfully running Sulu you'll have to fulfill the following system requir
 
 * Mac OSX, Linux or Windows
 * Webserver (`apache <http://httpd.apache.org/>`_ or `nginx <http://nginx.org/>`_ with enabled URL rewriting)
-* `PHP <http://php.net/>`_ 5.5 or higher
+* `PHP <http://php.net/>`_ 5.5 or higher with php-intl and php-gd
```
danrot: `php-intl` is probably correct, but you could also use `imagemagick` instead of `php-gd` (what we usually do...). I do not know how we should handle that in here 😕

book/getting-started/system-requirements.rst
```diff
@@ -29,4 +29,4 @@ If you want to build parts of the system on your own, you will additionally need
 * `ruby <https://www.ruby-lang.org/en/>`_
 * `Compass <http://compass-style.org/>`_
 
-Hopefully you got your Server running, so you're ready to start over with Sulu.
\ No newline at end of file
+Hopefully you got your Server running, so you're ready to start over with Sulu.
```
danrot: Looks like we are missing a new line here 😃

book/getting-started/system-requirements.rst
wachterjohannes: maybe say `php-intl and one of image-library php-gd or php-imagemagick`?

danrot: `and one of the image-libraries php-gd or imagick` - yeah, sounds ok for me.

book/getting-started/system-requirements.rst
wachterjohannes: i have already a line on my machine. so i think we can ignore this.

wachterjohannes:
@danrot this is now ready to merge.

PR #221, opened 2016-05-06
Fix example navigation twig

alexander-schranz:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

The navigation twig extension need a depth when you want a subnavigation.

#### Why?

Currently their is no depth given.

### DIFF
--- a/book/creating-a-basic-website/creating-a-twig-template.rst
+++ b/book/creating-a-basic-website/creating-a-twig-template.rst
@@ -121,7 +121,7 @@ context.
     :linenos:
 
     <ul>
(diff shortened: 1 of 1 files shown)


wachterjohannes:
@alexander-schranz is this also mentioned in the twig function reference?

alexander-schranz:
@wachterjohannes yes the depth is listed as param: http://docs.sulu.io/en/latest/reference/twig-extensions/functions/sulu_navigation_root_tree.html?highlight=sulu_navigation_root_tree

PR #240, opened 2016-07-25
adapted example configuration

danrot:
| Q | A
| --- | ---
| Fixed tickets | none
| License | MIT

#### What's in this PR?

This PR corresponds to the change in https://github.com/sulu/sulu/pull/2665.

### DIFF
--- a/book/getting-started/setup.rst
+++ b/book/getting-started/setup.rst
@@ -2,7 +2,7 @@ Setup a website!
 ================
 
 On this page you'll learn how to configure your website. You'll define webspaces
-and languages, create templates, set the correct permissions and create a user. 
+and languages, create templates, set the correct permissions and create a user.
 
 
 Webspaces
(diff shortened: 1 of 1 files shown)


book/getting-started/setup.rst
```diff
@@ -95,14 +90,15 @@ in squared brackets in the following example:
                     </environment>
                     <environment type="dev">
                         <urls>
-                            <url>[url]</url>
-                            <url language="en" country="us">localhost</url>
+                            <url>[url]/{localization}</url>
```
wachterjohannes: this `[` should be `{`

wachterjohannes:
Else ACK

book/getting-started/setup.rst
danrot: Actually not... These are the placeholders that have to be replaced by the user. We should actually think about updating this guide to use sulu-minimal, but that's a different story.

PR #242, opened 2016-08-01
WIP: added language import doc

dominikmatt:
| Q | A |
| --- | --- |
| SULU PR | sulu/sulu#2705 |


danrot:
Don't we have other example of commands? What you've described here you'll also get when simply typing the command in the command line, would add a bit more information, if we want to add that to the documentation.

PR #243, opened 2016-08-01
documented postgres option in test command

marcelmoosbrugger:
| Q | A
| --- | ---
| Fixed tickets | none
| License | MIT

#### What's in this PR?

A new option for the `runtest` command was introduced. This PR adds the documentation for it.

### DIFF
--- a/developer/contributing/test-your-code.rst
+++ b/developer/contributing/test-your-code.rst
@@ -37,6 +37,7 @@ the component tests (``C``).
    * ``-i``: Initialize the test setup (e.g. creating database).
    * ``-t [Bundle]``: Run the tests only for the specific bundle.
    * ``-a``: Run all tests.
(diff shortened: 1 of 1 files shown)


danrot:
The change on the documentation should actually not be necessary, because the environment configuration should be used for that (as described in http://docs.sulu.io/en/latest/developer/contributing/test-your-code.html?highlight=test#test-environment-configuration)

marcelmoosbrugger:
When setting the "-p" or "--postgres" option the runtest-script does exactly this. It starts the tests with some environment variables set to "postgres" or "pdo_pgsql". It's the same approach as with travis-CI. I found the flag very convenient, because as a developer you want to make sure that the tests are running. Therefore you just execute "bin/runtests -i -a" and "bin/runtests -i -a -p"

marcelmoosbrugger:
Obsolete

PR #256, opened 2016-08-17
removed the non-existing import command for translations

danrot:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

This PR removes the reference to the `sulu:translate:import` command.

#### Why?

Because it has been removed in https://github.com/sulu/sulu/pull/2608


danrot:
Replaced by #256

PR #266, opened 2016-08-31
Twig template example using wrong twig function

zelenay:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Fix example twig template of the smart content

#### Why?

There is content_path used in the example, but I thing that sulu_content_path should be used.


wachterjohannes:
@zelenay thanks for contribution

PR #275, opened 2016-09-27
Fixed Issue #272, sulu.rlp.part can be used multiple times (but without priority for now)

### LINKED ISSUE #272: sulu.rlp.part can be used multiple times
Can you add a description in the docs ...
http://docs.sulu.io/en/latest/reference/content-types/resource_locator.html
... that sulu.rlp.part can be used multiple times and the field values will be concatenated!?

Thanks!

baroso:
| Q | A
| --- | ---
| Fixed tickets | fixes #272 
| License | MIT

#### What's in this PR?

Explanation and code example.

#### Why?

Added documentation about sulu.rlp.part, that it can be used multiple times and the values are concatenated.

### DIFF
--- a/reference/content-types/resource_locator.rst
+++ b/reference/content-types/resource_locator.rst
@@ -19,10 +19,11 @@ Tags
     * - Tag
       - Description
     * - sulu.rlp
-      - The resource locator with this tag defines the URL to a specific page
+      - The resource locator with this tag defines the URL to a specific page.
     * - sulu.rlp.part
-      - Fields marked with this tag are used to generate the URL for a specific
-        page
+      - Fields marked with this tag are used to generate the URL for a specific page.
+        If more than one field ist marked, the values of these fields will be concatenated into the resource locator.
+        The sequence depends on the value of the priority parameter.
 
 Parameters
 ----------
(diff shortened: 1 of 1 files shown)


reference/content-types/resource_locator.rst
```diff
@@ -19,10 +19,11 @@ Tags
...
       - Description
     * - sulu.rlp
-      - The resource locator with this tag defines the URL to a specific page
+      - The resource locator with this tag defines the URL to a specific page.
     * - sulu.rlp.part
-      - Fields marked with this tag are used to generate the URL for a specific
-        page
+      - Fields marked with this tag are used to generate the URL for a specific page.
+        If more than one field ist marked, the values of these fields will be concatenated into the resource locator.
+        The sequence depends on the value of the priority parameter.
```
danrot: Sadly that's not true, this is a bug at the moment 😕 I think as long as it is not working we should not have it like this in the documentation.

danrot: Sadly that's not true, this is a bug at the moment 😕 I think as long as it is not working we should not have it like this in the documentation.

baroso: No problem, that's what I meant with my "should I wait" comment: https://github.com/sulu/sulu-docs/issues/272#issuecomment-248013989 ;)

danrot: Ah, ok, I was talking about something different 😅 We should change the documentation now, but omit the `priority` for the time being.

baroso: Did I do this right with my second commit?

danrot: Do it in the same PR. Ideally you would rebase the existing commit, but a separate commit is also fine.

PR #308, opened 2017-01-19
Added Docs for Snippet and Webspace import/export

dominikmatt:
| Q | A |
| --- | --- |
| SULU PR | sulu/sulu#2705 |


wachterjohannes:
@danrot technically seems to be ok but where did you think we should place this files? i dont see it in the book but also the cookbook seems not a very good choice!

danrot:
Definitely not the book, I would say this is a reference, no?

PR #316, opened 2017-02-25
Add example usage of countries parameter in location content type

mtnorthrop:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

This PR adds an example of how to use the `countries` parameter in the `location` content type.


wachterjohannes:
@mtnorthrop thanks for contribution.

PR #319, opened 2017-03-07
Add setup instruction for assetic bundle

revoltek-daniel:
#### What's in this PR?

Add missing Assetic Bundle install step if installation from min sulu setup project.

#### Why?
Setup themes correctly

### DIFF
--- a/book/themes.rst
+++ b/book/themes.rst
@@ -66,6 +66,36 @@ To enable it add the following lines into the `app/AbstractKernel.php` and
         load_controllers: false
         assetic_integration: true
 
+Make sure that the AsseticBundle is active, if not add
+
+.. code-block:: xml
+
+    abstract class AbstractKernel extends SuluKernel
+    {
+        /**
+         * {@inheritdoc}
+         */
+        public function registerBundles()
+        {
+            $bundles = [
+                ...
+                new Symfony\Bundle\AsseticBundle\AsseticBundle(),
+
+                ...
+            ];
+
+            ...
+
+            return $bundles;
+        }
+    }
+
+and run
+
+.. code-block:: xml
+
+    composer require symfony/assetic-bundle:~2.3
+
 This will configure a default theme which can be enabled in the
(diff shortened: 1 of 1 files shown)


book/themes.rst
```diff
@@ -66,6 +66,36 @@ To enable it add the following lines into the `app/AbstractKernel.php` and
         load_controllers: false
         assetic_integration: true
 
+Make sure that the AsseticBundle is active, if not add
+
+.. code-block:: xml
```
alexander-schranz: is php

alexander-schranz:
The assetic bundle itself is not needed for theming! You can use the theme bundle without the assetic bundle and just link in the different twig files to other resources.

Its only needed when use the `assetic_integration` or when you include the resources with assetic twig functions.

So in my case `assetic_integration: true` should not be required and the `Assetic` maybe an own section here how to include it. /cc @danrot @wachterjohannes

book/themes.rst
```diff
@@ -66,6 +66,36 @@ To enable it add the following lines into the `app/AbstractKernel.php` and
         load_controllers: false
         assetic_integration: true
 
+Make sure that the AsseticBundle is active, if not add
```
alexander-schranz: assetic bundle is not required when just using the theme bundle

wachterjohannes:
i would also add an own section - but i am not quite sure if we should handle this in our docu (which has also be maintained here) - when we dont rely on it @danrot what do you think?

revoltek-daniel:
So you should remove the ```assetic_integration: true``` part from the setup guide.

danrot:
What was the actual problem? Assetic should really not be required. Maybe only mention that the `assetic_integration` must be set to `true` when Assetic is used, but I would not explain how to install assetic. For that we can simply refer to the assetic documentation.

alexander-schranz:
assetic_integration has been removed from the docs some time ago to avoid the assitic not installed error.
As it also not recommended to use assetic anymore I will close this PR.

@revoltek-daniel still thank you for your contribution and show as an error in the documentation.

PR #343, opened 2017-06-12
added note for explaining how to configure doctrine repositories for the dataprovider

### LINKED ISSUE #322: Smart content provider missing set entity repository class in orm.xml
We should add a note or something to the http://docs.sulu.io/en/1.5/cookbook/smart-content-data-provider.html#service-definition that the Repository Class also need to be set on the entity.orm.xml file because Doctrine ignores the class and use the class from the .orm.xml file when not set its still EntityRepository.

danrot:
| Q | A
| --- | ---
| Fixed tickets | fixes #322
| License | MIT

#### What's in this PR?

Adds an explanation on how to correctly create repositories for usage with the SmartContent DataProvider.

### DIFF
--- a/cookbook/smart-content-data-provider.rst
+++ b/cookbook/smart-content-data-provider.rst
@@ -302,6 +302,13 @@ definition.
 Afterwards you can use your new DataProvider within a normal SmartContent
 property.
 
+.. note::
+
+    Mind that the `class` property should set to a sensible value, but it has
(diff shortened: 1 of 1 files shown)


danrot:
@alexander-schranz Do you think that explanation is enough regarding your issue #322?

alexander-schranz:
@danrot perfect 👍

PR #365, opened 2018-02-28
Update extend-entities.rst

cm-springlane:
Add the correct table name and a note in which file this sulu_security block belongs in

| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Add the correct table name and a note in which file this sulu_security block belongs in

#### Why?

Without the correct table name the wrong table was used.


wachterjohannes:
@cm-springlane thanks for this patch :)

PR #370, opened 2018-04-12
Missing type on Custom Tags

plozmun:
| Q | A
| --- | ---
| Fixed tickets | -
| License | MIT

#### What's in this PR?

Fix doc on Custom Tags [MarkupBundle]

#### Why?

Without this tag :
```
In TagCompilerPass.php line 46:
Notice: Undefined index: type  
```


wachterjohannes:
@plozmun thanks for fixing this :)

PR #374, opened 2018-04-19
Update image-formats.rst

marcel-burkhard:
use Version 1.1 of XSD

| Q | A
| --- | ---
| Fixed tickets | none
| License | MIT

#### What's in this PR?

Inspired by: https://github.com/sulu/sulu-demo/blob/master/app/config/image-formats.xml

#### Why?

I installed sulu yesterday and the old config format documented in http://docs.sulu.io/en/latest/book/image-formats.html doesn't seem to do anything, when I use the new one it works.


marcel-burkhard:
I forgot to remove the name node, I'll make a new one

PR #391, opened 2018-09-01
Fix sphinx build and errors in code highlighting

alexander-schranz:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Fix sphinx build in 1.7 and errors in code highlighting.

#### Why?

The sidebar is currently not working correctly because of bc breaks in the basic theme in sphinx 1.7.


_themes/kr/theme.conf
```diff
@@ -2,7 +2,8 @@
 inherit = basic
 stylesheet = flasky.css
 pygments_style = flask_theme_support.FlaskyStyle
+sidebars =
```
alexander-schranz: this was mostly the problem with the build they added new sidebars to the basic theme here: https://github.com/sphinx-doc/sphinx/commit/b78f3090a05f38fb3beb14f7e716f1b26a2ff253#diff-8157ebb8774f8bad9cb7f2c97b2b46ec
setting it to empty seems to fix it

PR #394, opened 2018-10-16
Update templates.rst

zangai:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Added properties node to clarify a block node needs to be nested in the properties node of the template instead of the template node.

#### Why?

The documentation falsely implied a block node should be put in the template node


alexander-schranz:
@zangai thank you for fixing this!

zangai:
@alexander-schranz You're welcome :)

PR #400, opened 2018-10-25
Start PHP server in dev environment.

ADmad:


wachterjohannes:
@ADmad thanks for this improvement

PR #401, opened 2018-11-05
Fixed type designator for selecting contacts

zangai:
The type with the designator "contact_selection" has been renamed to "contact"

| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

The type with the designator "contact_selection" has been renamed to "contact"

#### Why?

The docs stated it was contact_selection, but that doesn't work.


zangai:
Hmm, seems like it chokes on the reference. I don't know know that markup, so maybe anyone else can fix it? Information still applies.

wachterjohannes:
@zangai thanks for this contribution

PR #411, opened 2018-11-29
Add docs for multilingual="false"

lastzero:
| Q | A
| --- | ---
| Fixed tickets | not a fix, added info
| License | MIT

#### What's in this PR?

Docs for the `multilingual="false"` attribute. It is currently undocumented, but very important for many users. You can find it like that in unit tests and this is also the info we got from Patrick in the Slack chat (#general channel).

However - when I tested it today - it didn't seem to work with existing content (Sulu CMS 1.6.22). You probably need to clear/rebuild a cache? A core developer should add additional / missing information for this case.

#### Why?

Missing documentation for this attribute.

Otherwise you need to manually sync those properties between languages, if content must be the same for all languages (language independent/not multilingual). Our team spent multiple days in total doing this and it was not clear how to change it or if it can be changed at all.


danrot:
@lastzero Thanks for your contribution! But I think you have not chosen the best place for that... The page you have edited doesn't say anything about `properties`, so it might confuse some people... I think the best place would be as a sibling header to [Mandatory/Optional Properties](http://docs.sulu.io/en/latest/book/templates.html#mandatory-optional-properties). Could you possibly move that part there?

And regarding your question: I think it is not possible to have this without some kind of migration. If you have already data in multiple languages, which one should be taken if you decide to make it a non-multilingual property? And keeping it the way it is currently has another (dis)advantage (depends on how you look at it): It allows you to go back to the non-multilingual property, and you will still have your old data there :slightly_smiling_face:

And sorry for the long response time :confused:

lastzero:
Well, it is the place we were looking for these docs. I'll put this on my todo for after I've done a commercial project to buy food and drinks. I've been working for free on open-source projects for the last couple of months without getting a single donation and I'm not even using Sulu CMS myself.

lastzero:
Moved docs to `templates.rst` as requested and added a note on property migration. Does that work for you?

book/templates.rst
```diff
@@ -295,6 +295,38 @@ property, set the attribute ``mandatory`` to ``true``:
...
+                <!-- ... -->
+            </property>
+
+            <!-- ... -->
+        </properties>
+    </template>
+
+.. note::
+
+    If you change an existing attribute from a multilingual to a non multilingual version,
```
danrot: Would use `property` instead of `attribute here to stay consistent.

book/templates.rst
```diff
@@ -295,6 +295,38 @@ property, set the attribute ``mandatory`` to ``true``:
...
+    it is necessary to migrate the values from one language to the new attribute,
```
danrot: Also use `property` instead of `attribute`. And I would mention `the values in PHPCR` or something like that, since that is not completely clear.

danrot:
@lastzero Thanks for the contribution!

PR #413, opened 2018-12-08
Fix image download for nginx

alexander-schranz:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Fix image download for nginx.

#### Why?

Currently the fallback to index.php is not working.


danrot:
Makes sense to me, can somebody with a setup nginx test it? @trickreich @nnatter @wachterjohannes maybe one of you?

alexander-schranz:
ping @wachterjohannes

PR #427, opened 2019-02-14
Add example for removing all cookies still cache page in varnish vcl

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | 
| License | MIT

#### What's in this PR?

Add example for removing all cookies still cache page in the varnish vcl

#### Why?

In some cases where you example use the sulu community bundle you have sessions on your site and still want to cache this pages.

### DIFF
--- a/cookbook/caching-with-varnish.rst
+++ b/cookbook/caching-with-varnish.rst
@@ -148,6 +148,33 @@ The following will add full caching support for Sulu:
     sub vcl_recv {
         call sulu_recv;
 
+        // # Add a Surrogate-Capability header to announce ESI support.
+        // set req.http.Surrogate-Capability = "abc=ESI/1.0";
+
+        // # Remove all cookies except the session ID (SULUSESSID)
+        // # Check configured session ID name in your config/packages/framework.yaml
+        // if (req.http.Cookie) {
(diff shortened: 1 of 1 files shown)


cookbook/caching-with-varnish.rst
```diff
@@ -148,6 +148,33 @@ The following will add full caching support for Sulu:
...
+        //         # If there are no more cookies, remove the header to get page cached.
+        //         unset req.http.Cookie;
+        //     }
+        // }
+
+        // if (req.method != "GET" && req.method != "HEAD") {
+        //     return (pass);
+        // }
+
+        // # In different to the builtin.vcl of varnish we cache the page still when it have a cookie
```
wachterjohannes, line 173: i think this comment is missleading - the following statement says that the request will not be cached when a Authorization header isset.

wachterjohannes (changes requested):

cookbook/caching-with-varnish.rst
alexander-schranz: Ok that make we know think that this should not be the case but I'm not so familiar with varnish /cc @chirimoya is this config false or should that be the case?

alexander-schranz: fixed this based on our projects configuration

PR #433, opened 2019-03-26
Rename internal link content types to page selection

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | 
| License | MIT

#### What's in this PR?

Rename internal link content types to page selection.

#### Why?

Normalize content type names

### DIFF
--- a/reference/content-types/index.rst
+++ b/reference/content-types/index.rst
@@ -58,7 +58,7 @@ content type:
     color
     date
     email
-    internal_links
+    page_selection
     location
     media_selection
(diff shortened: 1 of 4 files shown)


reference/content-types/index.rst
```diff
@@ -58,15 +58,15 @@ content type:
     color
     date
     email
-    internal_links
+    page_selection
```
danrot: I think we've ordered the names alphabetically here, so that readers can easier find what they are looking for.

PR #445, opened 2019-08-23
Update installation instructions for AudienceTargetingBundle

nnatter:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

This PR updates the installation instructions for the `AudienceTargetingBundle`.

### DIFF
--- a/bundles/audience_targeting.rst
+++ b/bundles/audience_targeting.rst
@@ -26,6 +26,7 @@ file (`config/routes/sulu_admin.yaml`):
 
     # ...
     sulu_audience_targeting_api:
+        type: rest
         resource: "@SuluAudienceTargetingBundle/Resources/config/routing_api.yml"
         prefix: /admin/api
 
--- a/bundles/audience_targeting.rst
+++ b/bundles/audience_targeting.rst
@@ -12,42 +12,25 @@ target group of the visitor.
 Installation
 ------------
 
-Enable the bundle in `app/AbstractKernel.php`:
+Enable the bundle in `config/bundles.php`:
 
 .. code-block:: php
 
     <?php
-    abstract class AbstractKernel extends SuluKernel
-    {
-        // ...
-
-        public function reigsterBundles()
-        {
-            $bundles = [
-                // ...
-                new Sulu\Bundle\AudienceTargetingBundle\SuluAudienceTargetingBundle(),
-            ];
-
-            return $bundles;
-        }
-    }
+    Sulu\Bundle\AudienceTargetingBundle\SuluAudienceTargetingBundle::class => ['all' => true],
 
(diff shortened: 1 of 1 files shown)


bundles/audience_targeting.rst
```diff
@@ -12,42 +12,25 @@ target group of the visitor.
...
-            ];
-
-            return $bundles;
-        }
-    }
+    Sulu\Bundle\AudienceTargetingBundle\SuluAudienceTargetingBundle::class => ['all' => true],
 
 Add the routes for the administration interface to the routing configuration
-file (`app/config/admin/routing.yml`):
+file (`config/routes/sulu_admin.yaml`):
```
alexander-schranz: if we want to provide a recipe for this bundle we should move this routes into an own config file e.g: `sulu_audience_targeting_admin.yaml`

bundles/audience_targeting.rst
```diff
@@ -12,42 +12,25 @@ target group of the visitor.
...
-        resource: "@SuluAudienceTargetingBundle/Resources/config/routing.xml"
-        prefix: /admin/target-groups
-
     sulu_audience_targeting_api:
         resource: "@SuluAudienceTargetingBundle/Resources/config/routing_api.xml"
         prefix: /admin/api
 
 And the routes for the website in the corresponding configuration file
-(`app/config/website/routing.yml`):
+(`config/routes/sulu_website.yaml`):
```
alexander-schranz: same as above would add `sulu_audience_targeting_website.yaml`

bundles/audience_targeting.rst
danrot: I don't know if I would do that now.

1. If we do that, then the entire `sulu_admin.yaml` file should be gone.
2. If we support Symfony Flex, then these steps would be gone completely anyway.

alexander-schranz: correct we have all other routes in the sulu_admin.yaml as well

bundles/audience_targeting.rst
```diff
@@ -12,62 +12,48 @@ target group of the visitor.
...
 .. code-block:: yaml
 
     # ...
-    sulu_audience_targeting:
-        resource: "@SuluAudienceTargetingBundle/Resources/config/routing.xml"
-        prefix: /admin/target-groups
-
     sulu_audience_targeting_api:
-        resource: "@SuluAudienceTargetingBundle/Resources/config/routing_api.xml"
+        resource: "@SuluAudienceTargetingBundle/Resources/config/routing_api.yml"
```
alexander-schranz: @nnatter is here `type: rest` missing? for be compatible with symfony 3.4 name_prefix?

PR #449, opened 2019-09-25
Fix default-type attribute value in block.rst

CVogel:
fixed default-type attribute value to match example type name


alexander-schranz:
@CVogel thank you!

PR #452, opened 2019-10-04
Fix bundle documentation for Sulu 2.0

danrot:
| Q | A
| --- | ---
| Fixed tickets | ---
| License | MIT

#### What's in this PR?

This PR fixes the `bundles` directory based on the adjustments in Sulu 2.0.

#### Why?

Because the documentation should be working for our new version as well.


bundles/collaboration.rst
```diff
@@ -2,40 +2,16 @@ CollaborationBundle
...
-what we usually want, since no id indicates that a new record is created and
-there will not be a conflict if two users are adding new records at the same
-time.
-
-The return value of the method is used by a JavaScript hook, which starts the
-collaboration component with these options.
-
+other content managers are currently editing the same record. AJAX polling has
+been implemented, in order to notify current editors when other editors join
+editing the page. If websockets are not activated the first editor opening the
```
nnatter: `If websockets are not activated...` - do we still use websockets here?

danrot:
@nnatter Now it is finished :slightly_smiling_face:

bundles/http_cache.rst
```diff
@@ -108,24 +108,17 @@ Symfony Http Cache
 
 The Symfony HTTP cache is the default caching client for Sulu. It is integrated directly into Sulu.
 
-It works by "wrapping" the kernel. You can find it in the website front controller ``web/website.php``:
+It works by "wrapping" the kernel. This is done by the kernel itself, but in the front controller
```
nnatter: i think there is a negation missing?

danrot: Why? :thinking: I don't think so, although it is possible that the sentence could be rewritten to be easier to read. But I am currently not changing too much :see_no_evil:

alexander-schranz: @nnatter front controller = index.php. Front controller is the single entry point to a webapplicaiton, not be confused with symfony controllers

PR #455, opened 2019-10-09
added configuration hint to configure enum mapping_types for doctrine

bartrail:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Added a configuration hint how to add a mapping type to support enums (need for mysql 5.7.23)

#### Why?

Unable to run the installation command `bin/adminconsole sulu:build dev`

### DIFF
--- a/book/getting-started.rst
+++ b/book/getting-started.rst
@@ -95,6 +95,16 @@ Here is an example for using Sulu with MySQL:
 
     DATABASE_URL=mysql://db_user:db_password@127.0.0.1:3306/db_name
 
+You might need to adjust your ``config/packages/doctrine.yaml`` file to add a mapping for enums:
+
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@bartrail thank you for your PR but sulu doesn't use any enum doctrine type. Tested it on my mysql 5.7.24 and didn't got any problems. Where did the error come from do you have a stack trace? Did you add other bundles or created custom entities?

alexander-schranz:
I will close this PR. Still thank you for taking time to create one!
As discuss on slack with you it seems after you did rebuild your containers with mysql 5.7.24 it did work. As I really can not find a reference to enum in the codebase I would think more about a opcache or similiar issue. Else I would really think its a doctrine/dbal issue with that specific mysql version.

PR #459, opened 2019-10-28
Extend admin

danrot:
| Q | A
| --- | ---
| Fixed tickets | fixes ---
| License | MIT

#### What's in this PR?

This PR adds some documentation explaining how to add your own navigation, lists and forms to the Sulu administration interface, without touching any JS.

#### Why?

Because that is one of the best features in Sulu 2.0, and people need to have some documentation in order to use it.

#### ToDo

- [ ] Add some reference documentation for lists and form metadata
- [x] Avoid error on website because of `EventController`


book/extend-admin.rst
```diff
@@ -0,0 +1,553 @@
...
+First of all these controllers have to define standard REST actions. For a controller e.g. handling events, this means
+there has to be a `POST` action for creating events, a `PUT` action for modifying events, a `DELETE` action for
+deleting events and finally a `GET` action for retrieving information about an event. The `POST`, `PUT` and `GET`
+actions accept resp. return a JSON serialization of the entity it is handling, in our case an event. This serialization
+could e.g. look something like this:
+
+.. code-block:: json
+
+    {
+        "name": "Sulu Con 2020",
```
alexander-schranz: should contain an id or?

book/extend-admin.rst
```diff
@@ -0,0 +1,553 @@
...
+    use App\Entity\Event;
+    use FOS\RestBundle\Routing\ClassResourceInterface;
+    use Sulu\Component\Rest\ListBuilder\Doctrine\DoctrineListBuilderFactoryInterface;
+    use Sulu\Component\Rest\ListBuilder\Metadata\FieldDescriptorFactoryInterface;
+    use Sulu\Component\Rest\ListBuilder\PaginatedRepresentation;
+    use Sulu\Component\Rest\RestController;
+    use Sulu\Component\Rest\RestHelperInterface;
+    use Symfony\Component\HttpFoundation\Response;
+
+    class EventController extends RestController implements ClassResourceInterface
```
alexander-schranz: RestController is deprecated and should not be used.

book/extend-admin.rst
```diff
@@ -0,0 +1,816 @@
...
+that you can use other services when adding stuff to the administration interface.
+
+The two most important hooks are for views and navigation items.
+
+Views are `React`_ components, whereby Sulu comes with
+a few of them predefined. These predefined views can be configured via certain options, so that they are reusable in
+different contexts. Such a view takes most of the space of the screen, the only things being excluded from it being the
+toolbar on the very top of the screen and the navigation on the left.
+
+TODO: Add image further explaining the arrangement of screen space
```
danrot: fix this todo

book/extend-admin.rst
```diff
@@ -0,0 +1,816 @@
...
+        private $listBuilderFactory;
+
+        /**
+         * @var RestHelperInterface
+         */
+        private $restHelper;
+
+        public function __construct(
+            ViewHandlerInterface $viewHandler,
+            FieldDescriptorFactoryInterface $fieldDescriptorFactory,
```
danrot: This line causes the website not to work, because this controller will be registered, but the `FieldDescriptorFactoryInterface` does not exist. There are multiple solutions to this problem:

1. We add [this configuration from the workshop repo](https://github.com/sulu/sulu-workshop-symfony-live-berlin-2019/blob/607ac5b/config/services.yaml#L48) to the skeleton. Then we decide that all controllers for the Admin have to be located there, and that they cannot be used in a website context.
2. Move the `FieldDescriptorInterface` somewhere else. I could imagine to have a `SuluListBuilderBundle` or something like that containing all the `ListBuilder` stuff being registered in both the `admin` and `website` context.
3. Just make the default value of this line `null`, then Symfony won't complain, however, if the controller is still called there might be problems because this dependency is missing (which we either can hope that it just does not happen, or it has to be handled in this controller somehow).

chirimoya: @alexander-schranz @wachterjohannes @nnatter ???

chirimoya: 1 + 2 💪

book/extend-admin.rst
```diff
@@ -0,0 +1,822 @@
+Extend Admin UI
+===============
+
+The final section of the book has a deeper look at how to introduce your own navigation items, lists, forms and field
```
nnatter: `has a deeper look at how to integrate your own entities and views into the Sulu administration interface. This includes introducing new navigation entities, configuring new list views and configuring new form views.`

nnatter: would omit field types

alexander-schranz:
as 2.0.2 is released can we merged this @danrot ?

danrot:
@alexander-schranz No, the fix we need was not implemented in this release...

danrot:
Oh, completely forgot about https://github.com/sulu/skeleton/pull/19 :see_no_evil: We can indeed merge that, if everything's ok for you :sweat_smile:

PR #464, opened 2019-11-07
Remove writing plain HTML first part from twig documentation

danrot:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

This PR removes the advice that you should write plain HTML first, before actually writing the twig template.

#### Why?

I have talked to some developers in our company, and they don't do that anymore, because it does not really have any advantages. You can just write the HTML in the twig file directly, and there has never been a lot of stuff interfering.


danrot:
@sulu/core-team WDYT?

PR #478, opened 2020-01-09
Add  hint for urls in add localizations

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | -
| License | MIT

#### What's in this PR?

Add hint for urls in add localizations.

#### Why?

So it is known that you need also check the `urls` when adding new langauges.

### DIFF
--- a/book/localization.rst
+++ b/book/localization.rst
@@ -49,6 +49,11 @@ accepting new content. Afterwards nobody will have any permissions on this
 locale, so make sure that you add this permission in the permissions tab of
 the contacts.
 
+.. note::
+
+    When adding new localizations check the configuration of the ``urls`` in the 
+    environment part of the webspace which is described in :doc:`webspaces`.
+
 Adding custom localizations
 ---------------------------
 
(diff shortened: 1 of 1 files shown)


book/localization.rst
```diff
@@ -49,6 +49,11 @@ accepting new content. Afterwards nobody will have any permissions on this
 locale, so make sure that you add this permission in the permissions tab of
 the contacts.
 
+.. note::
+
+    When adding new localizations check the configuration of the ``urls`` in the
```
danrot: To check for what? This note confuses me more than it helps :thinking:

danrot: My guess would be that you are referring to the latest slack message, where only a single language was configured and therefore the other weren't accessible. Would describe that in a different manner, maybe something like this:

```
When adding a new localization you have to make sure that the localization is also covered by one of the webspaces URLs. This can either happen by using the `language` and `country` attributes of the `url` tag, or by specifying e.g. the `{localization}` placeholder. For more information see :doc:`webspace`.
```

And is there a way to directly jump to the "urls" anchor? :thinking:

alexander-schranz: @danrot couldn't find a way in rst to jump to a anchor just when creating a normal link referencing to the webspaces`.html` file.

PR #479, opened 2020-01-10
Add docs for default templates

cirykpopeye:
| Q | A
| --- | ---
| License | MIT
| PR | sulu/sulu#4963

#### What's in this PR?

Documentation on how to set Default Templates


danrot:
I would suggest adding this explanation in [Setup a webspace](http://docs.sulu.io/en/2.0/book/webspaces.html), because the rest of the XML file is also described there.

PR #481, opened 2020-01-14
Update Smart Content documentation for Sulu 2.0

marcostastny:
Fixed a place, where the old provider 'content' has not been replaced with 'pages' yet.

| Q | A
| --- | ---
| Fixed tickets | none
| License | MIT

#### What's in this PR?

Removed remaining old smart content provider and replaced with new one.

#### Why?

The provider 'content' has been replaced with the 'pages' provider in Sulu 2.0 but documentation is still showing old provider.

### DIFF
--- a/reference/content-types/smart_content.rst
+++ b/reference/content-types/smart_content.rst
@@ -33,7 +33,7 @@ Parameters
       - Description
     * - provider
       - string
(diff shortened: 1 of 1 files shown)


danrot:
That was the wrong branch :see_no_evil: Going to backmerge it.

PR #484, opened 2020-01-27
Update checkbox.rst

luca-rath:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Adds missing documentation for the `label` parameter of checkbox content type.

#### Why?

Because it's missing.


reference/content-types/checkbox.rst
```diff
@@ -21,6 +21,9 @@ Parameters
     * - default_value
       - string
       - Defines the default value of the checkbox. When not set the initial value of the checkbox is null.
+    * - label
+      -
```
danrot: This should probably say "string" as well :slightly_smiling_face:

PR #485, opened 2020-01-29
Add some extra info for the single_media_selection contentType

eekes:
The single_media_selection contentType was missing in the 'Creating a Page Template' page, also added links to the return value.

### DIFF
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -203,6 +203,9 @@ Here is a table with the content types shipped in Sulu core:
 | |media_selection|            | widget for selecting media (images,         | array containing arrays with            |
 |                              | documents)                                  | urls for every format                   |
 +------------------------------+---------------------------------------------+-----------------------------------------+
+| |single_media_selection|     | widget for a single media item (image,      | Media_ Object                           |
+|                              | document)                                   |                                         |
++------------------------------+---------------------------------------------+-----------------------------------------+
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@eekes Thank you for your Pull Request can you rebase your branch for the 2.0 release branch?

danrot:
I think I would not mention the `Media` object as you did it... Can we just remove these parts? The problem with this GitHub link, is that it potentially will link to the wrong version of the file once we release new versions.

eekes:
> I think I would not mention the `Media` object as you did it... Can we just remove these parts? The problem with this GitHub link, is that it potentially will link to the wrong version of the file once we release new versions.

@danrot Ok will do. The reason I did that was because I saw these pages were linking to the return types too:

https://docs.sulu.io/en/2.0/reference/content-types/single_contact_selection.html
https://docs.sulu.io/en/2.0/reference/content-types/single_account_selection.html

danrot:
That's interesting... But well, then you can leave it as it is :slightly_smiling_face: Have to think how to handle this on a more generic level in that case.

danrot:
But this still contains too many commits, can you change that? :slightly_smiling_face:

eekes:
@danrot I think it should be OK now... Thanks again @alexander-schranz for the rebasing info/help! Never did that before.

danrot:
Thank you very much for this contribution! :tada:

PR #506, opened 2020-04-17
Escape system column in SQL statement to make it compatible with MySQL 8.0

nnatter:


alexander-schranz:
should be also changed in the UPGRADE.md I think 🤔

PR #509, opened 2020-04-22
Highlight hint for adding permissions to user role after creating webspace

nnatter:
| Q | A
| --- | ---
| License | MIT

#### Why?

It feels like forgetting about this is a common mistake, Maybe highlighting the hint helps 🙂


alexander-schranz:
@nnatter merged also in 2.0 and master

PR #519, opened 2020-05-05
Update extend entities docs

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | fixes -
| License | MIT

#### What's in this PR?

Update extend entities docs.

#### Why?

The config file `app/config/config.yml` is not longer true.

### DIFF
--- a/cookbook/extend-entities.rst
+++ b/cookbook/extend-entities.rst
@@ -59,7 +59,7 @@ Create your own Entity for example in the `src` folder and extends the Entity wi
         /**
          * Get myProperty
          *
-         * @return string 
+         * @return string
          */
         public function getMyProperty()
         {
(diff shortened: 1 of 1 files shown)


cookbook/extend-entities.rst
```diff
@@ -76,11 +76,14 @@ Create your own Entity for example in the `src` folder and extends the Entity wi
 Configuration
 -------------
 
-You can specify your new entity and if it exists your repository in the `sulu_security` 
-configuration section in the file app/config/config.yml.
+You can specify your new entity and if it exists your repository the
```
danrot, line 79: "your repository in the configuration section of the file"

cookbook/extend-entities.rst
```diff
@@ -76,11 +76,14 @@ Create your own Entity for example in the `src` folder and extends the Entity wi
...
+configuration section in the file ``config/packages/*``.
```
danrot, line 80: Maybe we should also mention that these files have to be created if they don't exist? Because as far as I see not all of these files exist.

PR #520, opened 2020-05-07
Add warning that values expression of select is only evaluated on initial request

nnatter:
#### What's in this PR?

This PR adds a warning that the values expression of the `select` and `single_select` content type is only evaluated during the initial request.

#### Why?

Because this is a common misconception in our Slack channel 🙂

### DIFF
--- a/reference/content-types/select.rst
+++ b/reference/content-types/select.rst
@@ -64,7 +64,7 @@ You can use symfony expression language to access values from a service.
 
 .. note::
 
-    Be aware the the provided expression is only evaluated during the initial request to the administration interface.
+    Be aware that the provided expression is only evaluated during the initial request to the administration interface.
     If you want to provide a selection for your custom entity, you should configure the ``selection`` field-type as
     described in :doc:`../../book/extend-admin`.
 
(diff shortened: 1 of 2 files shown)


reference/content-types/select.rst
```diff
@@ -62,6 +62,12 @@ Example
 
 You can use symfony expression language to access values from a service.
 
+.. note::
+
+    Be aware the the provided expression is only evaluated during the initial request to the administration interface.
```
alexander-schranz, line 67: `the the`. And would highlight the ` ``expression`` `.

reference/content-types/single_select.rst
```diff
@@ -53,6 +53,12 @@ Example
 
 You can use symfony expression language to access values from a service.
 
+.. note::
+
+    Be aware the the provided expression is only evaluated during the initial request to the administration interface.
```
alexander-schranz, line 58: same as above:

`the the`. And would highlight the ` ``expression`` `.

reference/content-types/select.rst
nnatter: fixed the `the the` 👍 not sure about the `expression`, because it is not highlighted in the sentence above. but if you want, i can highlibht it 🙂

PR #528, opened 2020-06-18
Remove svg imagine adapter documentation

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | fixes -
| Related PRs | sulu/sulu#5327
| License | MIT

#### What's in this PR?

Update SVG documentation.

#### Why?

SVG Adapter not longer need to be installed manually.

### DIFF
--- a/bundles/media/imagine-adapter.rst
+++ b/bundles/media/imagine-adapter.rst
@@ -9,7 +9,6 @@ Available adapters are:
 * `gd`_
 * `imagick_extension`_
 * `vips_extension`_
-* `contao/imagine-svg`_
 
 GD
 --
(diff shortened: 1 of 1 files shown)


bundles/media/imagine-adapter.rst
```diff
@@ -61,16 +61,11 @@ To install it use following commands:
...
 
-This adapter only handles SVG images and is an addition to the other adapters.
-
-To install it use following command:
-
-.. code-block:: bash
-
-    composer require contao/imagine-svg
+The `contao/imagine-svg`_ adapter only handles SVG images and is an addition to the other adapters.
+It is shipped with sulu so it don't need to be installed manually.
```
danrot, line 68: `it doesn't`

danrot: And should we mention it at all, if it is not necessary to do anything? :thinking:

danrot: `SVG` is not really matching with the other headings on this level.

alexander-schranz: I can also remove it.

alexander-schranz: I removed it from the docs.

PR #539, opened 2020-07-13
Add documentation about webpack encore

luca-rath:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Add documentation about how to use Webpack Encore

#### Why?

Because it's asked a lot of times in the slack channel.

### DIFF
--- a/book/twig.rst
+++ b/book/twig.rst
@@ -196,5 +196,11 @@ this code:
 
 Image formats need to be defined in the `image_formats.xml`_ in your config.
 
-.. _image_formats.xml: https://github.com/sulu/sulu-standard/blob/master/src/Client/Bundle/WebsiteBundle/Resources/themes/default/config/image-formats.xml
+CSS / JS
+--------
+
+You can organize and build your website assets the way you are most familiar with.
+If you want to use Symfony's Webpack Encore, have a look at the :doc:`../cookbook/webpack-encore`.
+
+.. _image_formats.xml: https://github.com/sulu/skeleton/blob/master/config/image-formats.xml
 .. _Sulu Homepage: http://sulu.io
--- a/book/webpack-encore.rst
+++ b/book/webpack-encore.rst
@@ -1,0 +1,148 @@
+Using Webpack Encore for your website assets
+============================================
+
+Webpack Encore is a tool to manage your javascript and css assets. It
+seamlessly integrates webpack into your symfony application.
+
+Installation
+------------
+
+First add the `WebpackEncoreBundle`_ as a dependency to your
+composer.json file.
+
+.. code:: bash
+
+   composer require symfony/webpack-encore-bundle
+
+To enable it, add the following line to the ``config/bundles.php`` file,
+if that has not been done already for you by Symfony Flex:
+
+.. code:: php
+
+   return [
+       ...
+       Symfony\WebpackEncoreBundle\WebpackEncoreBundle::class => ['all' => true],
+   ];
+
+Configuration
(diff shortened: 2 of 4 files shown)


book/webpack-encore.rst
```diff
@@ -0,0 +1,148 @@
+Using Webpack Encore for your website assets
```
alexander-schranz, line 1: For me this should be more in the `cookbook` instead of `book`.

book/webpack-encore.rst
```diff
@@ -0,0 +1,148 @@
...
+use Webpack Encore with Sulu, some configuration has to be adjusted,
+because there is an additional build for Sulu’s admin interface.
+
+To continue, create a directory ``assets/website/`` and move the
+following directories and files into that directory:
+
+* ``assets/css/``
+* ``assets/js/``
+* ``package.json``
+* ``webpack.config.js``
```
alexander-schranz, line 52: I would not move this files as symfony recipes update for this files could be important and so I would keep them in the root folder of the project.

book/webpack-encore.rst
```diff
@@ -0,0 +1,148 @@
...
+* ``config/packages/assets.yaml``
+* ``config/packages/prod/webpack_encore.yaml``
+* ``config/packages/test/webpack_encore.yaml``
+* ``config/packages/webpack_encore.yaml``
+* ``package.json``
+* ``webpack.config.js``
+
+In a normal Symfony application, that would already work fine, but to
+use Webpack Encore with Sulu, some configuration has to be adjusted,
+because there is an additional build for Sulu’s admin interface.
```
nnatter, line 44: `there is an additional Javascript application for`

book/webpack-encore.rst
nnatter: 👍

book/webpack-encore.rst
luca-rath: I would rather add another section **Upgrading the Webpack Encore recipe**. Upgrading will never be straightforward, because flex will override all our changed files during the upgrade (all configuration files, app.js, app.css, package.json ...), so you would have to configure all of them again. @alexander-schranz @nnatter what do you think?

book/twig.rst
```diff
@@ -196,5 +196,11 @@ this code:
...
 Image formats need to be defined in the `image_formats.xml`_ in your config.
 
-.. _image_formats.xml: https://github.com/sulu/sulu-standard/blob/master/src/Client/Bundle/WebsiteBundle/Resources/themes/default/config/image-formats.xml
+CSS / JS
+--------
+
+You can organize and build your website assets the way you are most familiar with.
+If you want to use Symfony's Webpack Encore, have a look at the :doc:`../cookbook/webpack-encore`.
+
+.. _image_formats.xml: https://github.com/sulu/skeleton/blob/master/config/image-formats.xml
```
luca-rath, line 205: Updated this link

book/webpack-encore.rst
alexander-schranz: I would stay the symfony way as near as possible, don't see any advantages to force them that they need to go into the assets/website folder to run npm scripts for their application.

luca-rath: I'm not really happy with that, because I can't see a benefit in that change, but I implemented it as you wanted.

alexander-schranz: Its consistent for the developers which use symfony and sulu projects having the npm scripts for there application at the same places.

PR #540, opened 2020-07-17
Fix cache lifetime for javascript in apache vhost

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | fixes -
| License | MIT

#### What's in this PR?

It seems like some version of apache still return javascript as `application/javascript` instead of `text/javascript`.

E.g.: `curl -XGET "https://sulu.io/website/build/js/main.js?v=1.0.254" -I | grep 'Content-Type'`

#### Why?

That javascript files are not cached by the browser.


nnatter:
Should we target `1.6` here?

alexander-schranz:
Definitely just forgot to set the correct target branch.

PR #543, opened 2020-07-29
Fix controller naming for Symfony 5

wimvds:
| Q | A
| --- | ---
| Fixed tickets | 
| Related PRs | 
| License | MIT

#### What's in this PR?

A fix of the docs that is required for Symfony 5.

#### Why?

This PR fixes issues _Controller "SuluWebsiteBundle:Default:index" does neither exist as service nor as class_ errors when upgrading from 2.0 to Sulu 2.1 (the skeleton generates these correctly but the docs still use the old controller naming that no longer works in Symfony 5).

### DIFF
--- a/book/twig.rst
+++ b/book/twig.rst
@@ -19,7 +19,7 @@ In :doc:`templates` we learned how to define a template.
         <key>default</key>
 
         <view>templates/default</view>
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@wimvds Thank you!

PR #545, opened 2020-08-20
Add documentation about extending website search

luca-rath:
| Q | A
| --- | ---
| Related PRs | sulu/sulu#5460
| License | MIT

#### What's in this PR?

Add documentation about extending website search.

#### Why?

Because it should be documented.


cookbook/extending-website-search.rst
```diff
@@ -0,0 +1,35 @@
...
+The configuration looks like the following:
+
+.. code-block:: yaml
+
+    # config/packages/sulu_search.yaml
+
+    sulu_search:
+        website:
+            indexes:
+                articles: article_published
```
danrot, line 19: Shouldn't the articles be preprended by the `SuluArticleBundle`? Why should we burden the developer with doing this on their own?

danrot: Personally I would consider this a bug fix, not a BC break, therefore I would not care about the changed behavior.

alexander-schranz:
I personally worried that the index will not take the webspace settings into account and the search links will end in a 404?

@wachterjohannes or is this handled somewhere?

danrot:
Well, if that is the case we would have to fix that anyway, wouldn't we?

cookbook/extending-website-search.rst
```diff
@@ -0,0 +1,35 @@
+Extending website search
```
alexander-schranz, line 1: Not sure about the place should this not be in the `SuluSearchBundle`?

danrot: Good point, that file does even already exist: https://github.com/sulu/sulu-docs/blob/master/bundles/search.rst

wachterjohannes:
@alexander-schranz i am not sure that would have to be checked

alexander-schranz (approved):
Looks good from my side.

bundles/search.rst
```diff
@@ -53,6 +53,52 @@ So a sample configuration would look like this:
...
+An example for this would be the index for the pages:
+
+.. code-block:: yaml
+
+    # config/packages/sulu_search.yaml
+
+    sulu_search:
+        website:
+            indexes:
+                pages: page_#webspace#_published
```
danrot, line 100: Shouldn't we use another example? If people find that, they might think they have to do this, although I would say they absolutely should not... Maybe just use something like `example_#published#`?

danrot (approved):

PR #554, opened 2020-09-18
Change example from recipe to event

TheCadien:
#### What's in this PR?

For more uniformity the "Recipe" example has been changed to "Event

### DIFF
--- a/bundles/route/index.rst
+++ b/bundles/route/index.rst
@@ -131,7 +131,7 @@ EventController
     {
         public function indexAction(Event $event)
         {
-            return $this->render('App:website:event.html.twig', ['event' => $event]);
+            return $this->render('events/event.html.twig', ['event' => $event]);
         }
     }
 
@@ -158,7 +158,7 @@ RouteDefaultsProvider
         public function getByEntity($entityClass, $id, $locale, $object = null)
         {
             return [
-                '_controller' => 'App:Event:index',
+                '_controller' => 'App\Controller\EventController::indexAction',
                 'event' => $object ?: $this->eventRepository->find($id, $locale),
             ];
         }
@@ -182,7 +182,7 @@ simple call the `save` method of the service `sulu_route.manager.route_manager`.
 .. note::
 
     To update already existing entities you can run the command
-    `php bin/console sulu:route:update App:Event` which updates or creates
+    `php bin/console sulu:route:update App\\Entity\\Event` which updates or creates
     the route for all the entities of this type.
 
 Example without RoutableEntity
--- a/bundles/route/index.rst
+++ b/bundles/route/index.rst
@@ -36,7 +36,7 @@ Entity
 
 First extend your entity with the properties `id`, `locale` and a new many-to-one relation `route`:
 
-recipe.orm.xml:
+event.orm.xml:
 
 .. code-block:: xml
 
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@TheCadien Thank you, find this a great idea to use the same example in all cases. Can we rebase this to 2.0 branch would be great to be consistent through the whole 2.x documentation.

TheCadien:
@alexander-schranz sure i could do that

nnatter:
Hey, I rebased your changes onto the `2.0` branch. Hope thats okay 🙂

TheCadien:
@nnatter for sure. 
at you both. 
I checkt the the whole docs. The only other point where "recipes" is in used is the Teaser Selection. 
https://docs.sulu.io/en/2.0/reference/content-types/teaser_selection.html?highlight=teaser

In my mind it would be perfect too to change it in News. 
But in that Doc part are Images. Images which are realy helpfull i think. 

What do you think. Change on News and Remove the Images. ? 
Or anyone of you can create new images for "News ?
Or leave the area as it is.

cc @alexander-schranz

bundles/route/index.rst
```diff
@@ -140,26 +140,26 @@ RouteDefaultsProvider
...
-            $this->recipeRepository = $recipeRepository;
+            $this->eventRepository = $eventRepository;
         }
 
         public function getByEntity($entityClass, $id, $locale, $object = null)
         {
             return [
-                '_controller' => 'AppBundle:Recipe:index',
-                'recipe' => $object ?: $this->recipeRepository->find($id, $locale),
+                '_controller' => 'App:Event:index',
```
alexander-schranz, line 161: here we should go with FQCN as it is documented in the symfony docs for routing now:

```php
'_controller' => 'App\Controller\EventController::indexAction',
```

bundles/route/index.rst
```diff
@@ -107,31 +107,31 @@ Configure the route-schema in the file `app/config/config.yml`:
...
     use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 
-    class RecipeController extends AbstractController
+    class EventController extends AbstractController
     {
-        public function indexAction(Recipe $recipe)
+        public function indexAction(Event $event)
         {
-            return $this->render('AppBundle:website:recipe.html.twig', ['recipe' => $recipe]);
+            return $this->render('App:website:event.html.twig', ['event' => $event]);
```
alexander-schranz, line 134: Namespace should not be used here i would go here with something like:

```twig
events/event.html.twig
```

bundles/route/index.rst
```diff
@@ -182,7 +182,7 @@ simple call the `save` method of the service `sulu_route.manager.route_manager`.
 .. note::
 
     To update already existing entities you can run the command
-    `php bin/console sulu:route:update AppBundle:Recipe` which updates or creates
+    `php bin/console sulu:route:update App:Event` which updates or creates
```
alexander-schranz, line 185: Does this work? I think this also need to be the `FQCN` e.g.:

```bash
App\\Entity\\Event
```

alexander-schranz (changes requested):
Some small change request to be compatible to the symfony 4, 5 versions and twig 2 / 3.

alexander-schranz:
@TheCadien would see the teaser as an own PR and it should also use the `Event` entity. The event in the [sulu-workshop](https://github.com/sulu/sulu-workshop) has an image so I would just use the function which provided by that [entity](https://github.com/sulu/sulu-workshop/blob/064374757b041f96a5b08c8c441d86bbb1a4c87b/src/Entity/Event.php#L139-L142), but not focus on what you need todo to add a image to the entity, that is not part of the teaser documentation.

alexander-schranz:
Thank you @TheCadien

PR #560, opened 2020-09-24
Add documentation about https

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | fixes -
| Related PRs | -
| License | MIT

#### What's in this PR?

Add documentation about https.

#### Why?

Many times people run into issue that sulu generates `http` instead of `https` urls. I hope this documentation does help people to understand how symfony/sulu choose the protocol to generate the urls.

### DIFF
--- a/cookbook/https.rst
+++ b/cookbook/https.rst
@@ -1,7 +1,7 @@
 HTTPS
 =====
 
-Sulu generates URLs in the same way as Symfony does..
+Sulu generates URLs in the same way as Symfony does.
 If the requested protocol is ``https`` it will use automatically ``https`` to generate its URLs.
 
 If this does not work propertly, your setup probably includes some kind of proxy, e.g. a load balancer or a HTTP cache like Varnish.
--- a/cookbook/https.rst
+++ b/cookbook/https.rst
@@ -1,11 +1,11 @@
 HTTPS
 =====
 
-Sulu behave the same way as Symfony when generating urls.
-If the requested protocol is ``https`` it will use automatically ``https`` to generate its urls.
+Sulu generates URLs in the same way as Symfony does..
+If the requested protocol is ``https`` it will use automatically ``https`` to generate its URLs.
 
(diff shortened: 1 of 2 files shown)


cookbook/https.rst
```diff
@@ -0,0 +1,35 @@
+HTTPS
+=====
+
+Sulu behave the same way as Symfony when generating urls.
```
danrot, line 4: "Sulu generates URLs in the same way as Symfony does."

cookbook/https.rst
```diff
@@ -0,0 +1,35 @@
...
+If the requested protocol is ``https`` it will use automatically ``https`` to generate its urls.
```
danrot, line 5: `URLs`

cookbook/https.rst
```diff
@@ -0,0 +1,35 @@
...
+
+If this does not happen its a hint that you are behind a load balancer or have a proxy between your request server.
```
danrot, line 7: "If this does not work properly, that your setup includes some kind of proxy, e.g. a load balancer or a HTTP cache like Varnish."

cookbook/https.rst
```diff
@@ -0,0 +1,35 @@
...
+In that case you should have a look at the `Symfony Proxy Documentation`_ and configure your trusted proxy ip addresses.
```
danrot, line 8: "In that case you should add the IP address of your trusted proxy as explained in the `Symfony Proxy Documentation`_."

danrot (changes requested):

cookbook/https.rst
danrot: Yeah, that's right... But if at all it would be `it's a hint`. (its shows some kind of possession). But I think I would like it even better this way:

"If this does not work propertly, your setup probably includes some kind of proxy, e.g. a load balancer or a HTTP cache like Varnish."

alexander-schranz: @danrot fixed!

PR #561, opened 2020-09-28
Add documentation for image map content type

luca-rath:
| Q | A
| --- | ---
| Related PRs | sulu/sulu#5434
| License | MIT

#### What's in this PR?

Add documentation for image map content type.

#### Why?

Every content type should be documented.

### DIFF
--- a/reference/content-types/image_map.rst
+++ b/reference/content-types/image_map.rst
@@ -1,5 +1,5 @@
 ImageMap
-=====
+========
 
 Description
 -----------
--- a/reference/content-types/image_map.rst
+++ b/reference/content-types/image_map.rst
@@ -22,27 +22,145 @@ No parameters available
 Example
 -------
 
+XML
+^^^
+
 .. code-block:: xml
 
-    <property type="image_map" name="special_image" default-type="text">
+    <property name="image_map" type="image_map" default-type="basic">
         <meta>
-            <title lang="de">Spezial-Bild</title>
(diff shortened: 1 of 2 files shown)


nnatter:
can we add a basic twig example here?

luca-rath:
@nnatter Would it be enough to just link here? https://github.com/sulu/sulu/blob/master/templates/pages/includes/image_map.html.twig

nnatter:
I dont have a strong opinion 🙂  What do you think? @alexander-schranz

luca-rath:
Because this is not a "simple" example, I would prefer that to not bloat the documentation

luca-rath:
but it's still the most minimal I could think of

luca-rath:
@alexander-schranz whats your opinion on this?

alexander-schranz:
@luca-rath then its more important as we have an example in the example.html.twig you could just copy that into this file.

luca-rath:
@alexander-schranz I'm sorry, I think I don't know exactly what you mean now 🙂 Should I copy just the link to the example or the whole example itself?

alexander-schranz:
@luca-rath whole twig example code.

luca-rath:
@alexander-schranz changed it

alexander-schranz (approved):

alexander-schranz:
@luca-rath awesome. Thank yoU!

PR #564, opened 2020-09-30
Restructure and reword documentation about default snippet functionality

nnatter:


cookbook/default-snippets.rst
```diff
@@ -1,62 +1,64 @@
...
-for different snippet content types or load a snippet by its area key with
-the twig extension. The areas can be defined in the xml template of the snippets.
+This functionality is especially helpful to allow the content manager to managing
+webspace-wide data such as social media links or contact information.
+
+Defining possible areas for a snippet type
+------------------------------------------
+
+Each snippet template can define multiple snippet areas. All defined snippet areas will
+be displayed on the ``Snippet Area`` tab of the webspace in the administration interface
```
alexander-schranz, line 21: This tab should be something like  `Areas` or `Snippet Areas` but sadly currently here still the name `Default Snippets` is used.

nnatter: not really related with this PR, but why dont you like `Default Snippets`? 🙂

alexander-schranz: Because that wording should since introduciton of the areas not longer exist sulu/sulu#3402

PR #569, opened 2020-10-19
Add explanation for segment webspace configuration

danrot:
| Q | A
| --- | ---
| Fixed tickets | ---
| Related PRs | https://github.com/sulu/sulu/pull/5277
| License | MIT

#### What's in this PR?

This PR explains the segment tag for the webspace.

#### Why?

Because people need that information if they want to make use of segments.

### DIFF
--- a/book/webspaces.rst
+++ b/book/webspaces.rst
@@ -55,6 +55,21 @@ following paragraphs.
             </contexts>
         </navigation>
 
+        <segments>
+            <segment key="w" default="true">
+                <meta>
+                    <title lang="en">Winter</title>
+                    <title lang="de">Winter</title>
+                </meta>
+            </segment>
+            <segment key="s" default="false">
+                <meta>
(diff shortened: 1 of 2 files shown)


book/webspaces.rst
```diff
@@ -55,6 +55,21 @@ following paragraphs.
             </contexts>
         </navigation>
 
+        <segments>
```
alexander-schranz, line 58: I would not add `<segments>` XML Tag here to the basic webspace example we use here in the install example. As I think its an edge case and we also do not added `<security>` and `<theme>` here in the basic book example.

danrot: I think that this is mixed here. If we really want to keep it minimal here, we would probably also have to remove the `excluded-templates` and `resource-locator` tags. But they are also here and only marked as "optional" in their headings. And since there is currently no other place where we are explaining the Webspace XML file I don't really see where else we should put it.

PR #571, opened 2020-10-23
Add @RouteResource annotation in Extend Admin UI example

nnatter:
| Q | A
| --- | ---
| License | MIT

#### Why?

When using an entity name that consists of two words (eg. `EventRegistration`), the `FOSRestBundle` will position the `id` parameter in the path between the `event` and `registration`: `/admin/api/events/{id}/registrations.{_format} `

This is not compatible with the Sulu UI. The Sulu UI expects the `id` parameter at the end of the path, because it needs to be omitted for the `POST` request when creating a new entity.

### DIFF
--- a/book/extend-admin.rst
+++ b/book/extend-admin.rst
@@ -211,7 +211,7 @@ shows a controller doing what has just been described.
     use Symfony\Component\HttpFoundation\Response;
 
     /**
-     * @RouteResource("Event")
+     * @RouteResource(Event::RESOURCE_KEY)
      */
     class EventController implements ClassResourceInterface
     {
--- a/book/extend-admin.rst
+++ b/book/extend-admin.rst
@@ -211,7 +211,7 @@ shows a controller doing what has just been described.
     use Symfony\Component\HttpFoundation\Response;
 
     /**
-     * @RouteResource(Event::RESOURCE_KEY)
+     * @RouteResource("event")
      */
     class EventController implements ClassResourceInterface
     {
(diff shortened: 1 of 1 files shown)


book/extend-admin.rst
```diff
@@ -209,6 +210,9 @@ shows a controller doing what has just been described.
     use Sulu\Component\Rest\RestHelperInterface;
     use Symfony\Component\HttpFoundation\Response;
 
+    /**
+     * @RouteResource("Event")
```
alexander-schranz, line 214: I think this should not be uppercase:

```
@RouteResource("event")
```

e.g.:

```
@RouteResource("event-registration")
```

alexander-schranz (changes requested):

book/extend-admin.rst
nnatter: the `FOSRestBundle` will lower case this automatically. if you dont add any annotation, the bundle will use `Event` (capitalized) too: https://github.com/handcraftedinthealps/RestRoutingBundle/blob/a9a11287e5b04235c18b1b46ab8dc69408191892/Routing/Loader/Reader/RestControllerReader.php#L78-L98

alexander-schranz: yeah but it would be better represents that its a resource key if you set it to lowercase

alexander-schranz: What happens if we use the resource key here does it work correctly?

```
@RouteResource(Event::RESOURCE_KEY)
```

nnatter: yes, looks like that is working 🙂

nnatter: but that will lead to the same problem if you use a `_` in your resource key 🙈

alexander-schranz: 🙈

PR #594, opened 2020-12-02
Add enableSorting call in DataProvider cookbook

nnatter:


cookbook/smart-content-data-provider.rst
```diff
@@ -236,13 +236,16 @@ to avoid filtering for these values.
         {
             if (!$this->configuration) {
                 $this->configuration = self::createConfigurationBuilder()
-                    ->enableTags()
                     ->enableLimit()
                     ->enablePagination()
                     ->enablePresentAs()
+                    ->enableSorting([
+                        ['column' => 'column', 'title' => 'sort-translation-key-1'],
```
Prokyonn, line 243: I'm not sure if this is not confusing with `column` as value. I would go for `id` or another concrete column name for the first entry. The second entry should make it clear, that you need to insert the database column/relation.

PR #597, opened 2020-12-06
Remove display_options param from smart_content documentation

nnatter:
It looks like this param is not handled in the frontend at all when using Sulu 2 😕

### DIFF
--- a/reference/content-types/smart_content.rst
+++ b/reference/content-types/smart_content.rst
@@ -68,10 +68,6 @@ Parameters
     * - category_root
       - string
       - Root category (key) to display category-tree.
-    * - display_options
-      - collection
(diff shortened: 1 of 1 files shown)


reference/content-types/smart_content.rst
```diff
@@ -68,10 +68,6 @@ Parameters
     * - category_root
       - string
       - Root category (key) to display category-tree.
-    * - display_options
```
alexander-schranz, line 71: /cc @chirimoya @danrot is this something we forget in 2.0 or was it also not possible in 1.6?

danrot: TBH I don't know, would have to test it myself...

PR #611, opened 2020-12-21
Emphasize the possibility of using nested blocks since Sulu 2.1

S1ructure:
| Q | A
| License | MIT

What's in this PR?
Updating the documentation about the possibility using nested blocks

Why?
Related to https://sulu.io/blog/sulu-2-1-rc1-is-here-with-symfony-5-support-nested-blocks-and-deep-links and related to my own tests sulu supports nesting starting with version 2.1.


nnatter:
I think we can just remove that sentence now. I already adjusted this PR. Thanks a lot for your time! 🙂

Issue #460, opened 2019-10-30
Missing docs about snippets


alexander-schranz:
There should be a single documentation about snippets and how to use them

 - What are snippets
 - It should show how to create snippets
 - How to embed snippets using the snippet selection.
 - How to define snippet areas
   - How to use area in snippet selection
   - How to use sulu_snippet_load_by_area

PR #613, opened 2020-12-22
Add simple snippets doc

TheCadien:
| Q | A
| --- | ---
| Fixed tickets | #460
| License | MIT

#### What's in this PR?

Add simple Snippet Doc

#### Why?

Doc is missing

### DIFF
--- a/book/index.rst
+++ b/book/index.rst
@@ -16,7 +16,6 @@ Be sure to know a thing or two about `Symfony`_. Sulu is based on Symfony and st
     twig
     content-architecture
     webspaces
-    snippets
     image-formats
     localization
     smart-content
--- a/book/index.rst
+++ b/book/index.rst
@@ -16,6 +16,7 @@ Be sure to know a thing or two about `Symfony`_. Sulu is based on Symfony and st
     twig
     content-architecture
     webspaces
+    snippets
     image-formats
     localization
     smart-content
--- a/book/snippets.rst
+++ b/book/snippets.rst
@@ -72,24 +72,3 @@ Properties
 ----------
 
 Properties are the same as Page :doc:`templates`.
-
-Load Snippets from a Subfolder
-------------------------------
-
-As already mentioned, Sulu loads snippet templates in its default configuration from the following path.
-`config/template/snippets/`
-
-By the means of configuration in `config/packages/sulu_admin.yaml` according to the following scheme
-it is also possible to load snippet templates from custom folders.
-
-.. code-block:: yaml
-
-    sulu_core:
-        content:
-            structure:
-                paths:
-                    events:
-                        path: "%kernel.project_dir%/config/template/events/snippets/"
-                        type: "snippet"
-
-In this example, a new Events folder has been specified. It is important that the key for the configuration remains unique for each config.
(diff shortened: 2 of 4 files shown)


book/snippets.rst
```diff
@@ -0,0 +1,95 @@
...
+            </property>
+        </properties>
+    </template>
+
+Properties
+----------
+
+Properties are the same as Page :doc:`templates`.
+
+Load Snippets from a Subfolder
```
alexander-schranz: As this is very uncommen to be changed. This should be moved into an own PR and should be added here about documenting how to change `page` and `snippet` folder directories:

https://docs.sulu.io/en/2.0/cookbook/override-dir-structure.html

book/snippets.rst
```diff
@@ -0,0 +1,95 @@
...
+By the means of configuration in `config/packages/sulu_admin.yaml` according to the following scheme
+it is also possible to load snippet templates from custom folders.
+
+.. code-block:: yaml
+
+    sulu_core:
+        content:
+            structure:
+                paths:
+                    events:
```
alexander-schranz: a key like `event_snippets` would make more sense to not conflict when somebody has an `events` entity.

book/snippets.rst
TheCadien: ah, haven't seen the page at all.  That makes sense, I remove it and add it there again.

book/snippets.rst
TheCadien: Change it at new pr

book/index.rst
```diff
@@ -16,6 +16,7 @@ Be sure to know a thing or two about `Symfony`_. Sulu is based on Symfony and st
     twig
     content-architecture
     webspaces
+    snippets
```
niklasnatter: I think I would not add an additional chapter in the book, because a lot of projects dont need to use snippets. Instead i would structure it like this:

1. Add a section below the `Content Blocks` section in the templates chapter: https://docs.sulu.io/en/2.2/book/templates.html#content-blocks
2. Shortly explain snippets and its usecases in the section and add a link to the`SnippetBundle` documentation
3. Create the `SnippetBundle` documentation (similar to the `LocationBundle`) https://docs.sulu.io/en/latest/bundles/location.html 

What do you think about this? 🙂

book/snippets.rst
```diff
@@ -0,0 +1,74 @@
+Using Snippets
```
niklasnatter: would use this file as `SnippetBundle` documentation (similar to `LocationBundle` or `RouteBundle`)

nnatter:
@TheCadien Like this - thanks a lot for your time! 🙂

book/index.rst
TheCadien: See your point. 

I generally wanted to expand on the snippets a bit more. 
However, I have decided for time reasons to create a first basic version as PR and add the rest after. 

Therefore the decision of a separate chapter.

but im interestet in your opinion, because of me, we can also gladly transfer the content to another chapter. :) !

niklasnatter: yeah, i think it would make sense to create a `SnippetBundle` documentation similar to the `LocationBundle` for this 🙂

TheCadien:
@nnatter I have changed the doc to the SnippetBundle :)

niklasnatter:
@TheCadien Thanks a lot! 🙂

PR #653, opened 2021-05-24
Fix image formats xml for gamma transformation

ternel:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Image-formats documentation: fix double `</format>` in example for gamma transformation.


alexander-schranz:
@ternel Thank you for the PR did rebase it to 2.1 so its also fixed on the old docs version.

alexander-schranz (approved):

PR #657, opened 2021-05-27
Add documentation about custom jexl binary operators AND and OR

luca-rath:
| Q | A
| --- | ---
| Fixed tickets | -
| Related PRs | sulu/sulu#6086
| License | MIT

#### What's in this PR?

Add documentation about custom jexl binary operators AND and OR

### DIFF
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -336,6 +336,11 @@ Inside of these attributes, you can utilize the `jexl`_ syntax for expressing yo
 
 .. note::
 
+    Because the `&` character cannot be used inside xml, you have to use `AND` instead of `&&` to
+    compare two conditions using a logical and. Additionally, you can also use `OR` instead of `||`.
+
+.. note::
+
     At the moment, conditions can only access the root properties of the template.
     This means that it is not possible to express conditions relative to a property inside a block.
 
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -336,7 +336,7 @@ Inside of these attributes, you can utilize the `jexl`_ syntax for expressing yo
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


book/templates.rst
```diff
@@ -334,6 +334,11 @@ Inside of these attributes, you can utilize the `jexl`_ syntax for expressing yo
         </properties>
     </template>
 
+.. note::
+
+    Because the `&` character cannot be used inside xml, you have to use `AND` instead of `&&` to
+    compare two conditions using a logical and. Additionally, you can also use `OR` instead of `||`.
```
niklasnatter:
```suggestion
    Because the ``&`` character cannot be used inside of XML files, you have to use ``AND`` instead of ``&&`` 
    if you want to to connect conditions using a logical and. Additionally, you can use ``OR`` instead of ``||``.
```

book/templates.rst
```diff
@@ -334,6 +334,11 @@ Inside of these attributes, you can utilize the `jexl`_ syntax for expressing yo
         </properties>
     </template>
 
+.. note::
+
+    Because the ``&`` character cannot be used inside of XML files, you have to use ``AND`` instead of ``&&``
```
alexander-schranz: > Because the ``&`` character cannot be used inside of XML files 

this is not technical correct would write as the `&` character need to be escaped in XML files, you can use ``AND`` instead of ``&&``

luca-rath:
```suggestion
    Because the ``&`` character needs to be escaped inside of XML files, you have to use ``AND`` instead of ``&&`` 
```

niklasnatter (approved):

PR #660, opened 2021-06-16
Add project root to cleanup step in build-admin documentation

niklasnatter:


niklasnatter:
This should not be necessary anymore after https://github.com/sulu/sulu/pull/6117 was merged.

PR #671, opened 2021-08-31
Extend activity bundle doc with translation example

TheCadien:
#### What's in this PR?

Extend the Activity Bundle Doc with simple example to get ressource title in the description and securing the list
And fix some small typo.

#### Why?
Missing doc


niklasnatter:
Thank you! 🙂

PR #673, opened 2021-09-10
Add documentation for single_collection_selection and collection_selection

niklasnatter:


reference/content-types/collection_selection.rst
```diff
@@ -0,0 +1,59 @@
+Collection selection
```
alexander-schranz, line 1: Actually they don't exist as content types as I know (there were some discussions but currently nothing which make sense to add them as content types)

As this is the documentation for page content types. We currently do not yet have a field type documentation, so we should discuss where we put field types and field type specific configs like (request_parameters) and where we put content types and content type specific configuraitons.

niklasnatter: i have created a PR to add the content types in https://github.com/sulu/sulu/pull/6219 because we need a `single_collection_selection` for https://github.com/sulu/sulu/pull/6218 🙂 

but i am open to only add the field-types but not the content-types if we dont want a `single_collection_selection`/`collection_selection` content type

alexander-schranz: @nnatter oh didn't saw that. Seems like the PR template was not used here 😄 . I'm in a conflict what somebody would expect by such a content type. A collection, medias, multiple collection, medias from multiple collections, with children? We still need to decide what returned in the headless bundle (id, title, mediaCount?, collectionCount?, no medias as I think smart content is for that).

If we go with the current implemented way of the content type I would atleast link here to the media smart content provider, so if somebody search to get medias by a collection to make sure that that smart content is the correct way then.

niklasnatter: yes i think the output isnt as clear as for other content types here. i think i would keep things consistent with the other selections and return the selected entities. if anybody needs something else, he could load it in a custom controller or a twig extension. i added a note about this and the `smart_content` to the documentation 🙂

alexander-schranz (approved):

PR #682, opened 2021-10-19
Add documentation: creating a document via a command

szz:
history: https://sulu-io.slack.com/archives/C0430GGCV/p1634221207268000

| Q | A
| --- | ---
| Fixed tickets | 
| Related PRs | 
| License | MIT

#### What's in this PR?

Add documentation: creating a document via a command

### DIFF
--- a/reference/components/document-manager/using_the_document_manager.rst
+++ b/reference/components/document-manager/using_the_document_manager.rst
@@ -62,6 +62,14 @@ on which subscribers you have registered.
 
 See the :doc:`subscribers` chapter for more information.
 
+.. note::
+
+    To create a document via a Sulu CLI command (e.g., import from an external source), run the command within the admin context.
+
+    .. code-block:: bash
+
+        bin/adminconsole your:command
+
 The Path Builder
 ----------------
 
--- a/reference/components/document-manager/using_the_document_manager.rst
+++ b/reference/components/document-manager/using_the_document_manager.rst
@@ -64,7 +64,7 @@ See the :doc:`subscribers` chapter for more information.
 
 .. note::
 
-    To create a document via a Sulu CLI command (e.g., import from an external source), run the command within the admin context.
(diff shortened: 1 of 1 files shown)


reference/components/document-manager/using_the_document_manager.rst
```diff
@@ -62,6 +62,14 @@ on which subscribers you have registered.
 
 See the :doc:`subscribers` chapter for more information.
 
+.. note::
+
+    To create a document via a Sulu CLI command (e.g., import from an external source), run the command within the admin context.
+
+    .. code-block:: bash
+
+        bin/adminconsole your:command
```
alexander-schranz: This text should be more general as it need to be the `admin` context. So you can not create a document inside a website controller you will need.

So something like "Documents can only be created in the admin context. For cli this means using the the `bin/adminconsole` and for controller it is required to have run them in the `/admin` context or using a symfony messenger to create them asynchron via admin CLI."

alexander-schranz (changes requested):

alexander-schranz (changes requested):
@szz Thank youf or the pull request! As this is not only for `2.x` we should target here atleast the `2.3` branch which is the current lowest supported version of sulu 2.

reference/components/document-manager/using_the_document_manager.rst
Prokyonn:
```suggestion
    Be aware that documents can only be created in the admin context, not in the website context. For CLI commands, this means you must use ﻿``bin/adminconsole`` to execute the command. Alternatively, you can execute a command that dispatches messages via the Symfony Messenger. In this case, the consumer that consumes the dispatched message must also run in the admin context.

    .. code-block:: bash

        bin/adminconsole your:command
```

Prokyonn (approved):
@alexander-schranz Please verify the adapted text :)

alexander-schranz (approved):

alexander-schranz:
@szz thank you!

PR #697, opened 2022-02-04
Add Tip for Tags in TrashItemHandler

TheCadien:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Add a Tip for Tagging the TrashItemHandler in External Services

#### Why?

Missing Docs

### DIFF
--- a/bundles/trash.rst
+++ b/bundles/trash.rst
@@ -117,6 +117,13 @@ with the correct data.
         }
     }
 
+.. Tip::
+
+    If `TrashItemHandler` is created outside of a Sulu sulu/skeleton application, for example in a separate bundle, the automatic autowire of the Trash bundle does not work.
+    In this case the handler must be tagged with the following tags.
+    ``sulu_trash.store_trash_item_handler`` , ``sulu_trash.restore_trash_item_handler`` and ``sulu_trash.restore_configuration_provider``.
+
+
 (Optional) Add restore configuration for your entity
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/bundles/trash.rst
+++ b/bundles/trash.rst
@@ -74,6 +74,12 @@ in time.
         }
     }
 
+.. Tip::
+
+    If you're using the default `services.yaml configuration from Symfony`_, the example should work as is.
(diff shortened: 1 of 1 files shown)


bundles/trash.rst
```diff
@@ -117,6 +117,13 @@ with the correct data.
         }
     }
 
+.. Tip::
+
+    If `TrashItemHandler` is created outside of a Sulu sulu/skeleton application, for example in a separate bundle, the automatic autowire of the Trash bundle does not work.
+    In this case the handler must be tagged with the following tags.
+    ``sulu_trash.store_trash_item_handler`` , ``sulu_trash.restore_trash_item_handler`` and ``sulu_trash.restore_configuration_provider``.
```
alexander-schranz: We should add a tip with the specific tag after every code block and not only to one of the blocks:

 - https://docs.sulu.io/en/2.4/bundles/trash.html#create-a-trash-item-for-an-entity
 - https://docs.sulu.io/en/2.4/bundles/trash.html#restore-an-entity-from-a-trash-item
 - https://docs.sulu.io/en/2.4/bundles/trash.html#optional-add-restore-configuration-for-your-entity

bundles/trash.rst
```diff
@@ -117,6 +117,13 @@ with the correct data.
         }
     }
 
+.. Tip::
+
+    If `TrashItemHandler` is created outside of a Sulu sulu/skeleton application, for example in a separate bundle, the automatic autowire of the Trash bundle does not work.
```
alexander-schranz: Would here go with the same links and wording like in the [symfony documentation](https://symfony.com/doc/5.4/form/create_custom_field_type.html):

> If you're using the [default services.yaml configuration](https://symfony.com/doc/5.4/service_container.html#service-container-services-load-example), this example will already work! Otherwise, [create a service](https://symfony.com/doc/5.4/service_container.html#service-container-creating-service) for this class and [tag it](https://symfony.com/doc/5.4/service_container/tags.html) with ``sulu_trash.store_trash_item_handler``.

alexander-schranz (changes requested):
Some small change request to split the tip and change the wording of the message a little bit to match symfony here better.

Prokyonn:
@alexander-schranz added your suggestions, please have a final look :)

alexander-schranz (approved):

alexander-schranz:
@TheCadien Thank you!

PR #704, opened 2022-04-10
Admin assets build when custom webpack config

oprichnik:
https://github.com/sulu/sulu-demo/issues/99

### DIFF
--- a/cookbook/webpack-encore.rst
+++ b/cookbook/webpack-encore.rst
@@ -140,6 +140,10 @@ and run the following command:
 
 Now you should be able to see the outcome in the browser.
 
+.. note::
+
+   When using custom assets with webpack-encore, you need to update the admin build in the public/build/admin directory: :doc:`build-admin-frontend`
+
 Optional: Install Web-JS
 ------------------------
 
--- a/cookbook/webpack-encore.rst
+++ b/cookbook/webpack-encore.rst
@@ -142,7 +142,8 @@ Now you should be able to see the outcome in the browser.
 
 .. note::
 
-   When using custom assets with webpack-encore, you need to update the admin build in the public/build/admin directory: :doc:`build-admin-frontend`
+   If webpack encore assets were accidentally build before changing the `public/build` to `public/website/build`
+   it could happen that all your admin assets were removed. Use `git checkout public/build/admin` or the `bin/console sulu:admin:update-build` command to restore the admin build files: :doc:`build-admin-frontend`
 
 Optional: Install Web-JS
 ------------------------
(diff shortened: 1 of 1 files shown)


cookbook/webpack-encore.rst
```diff
@@ -140,6 +140,10 @@ and run the following command:
 
 Now you should be able to see the outcome in the browser.
 
+.. note::
+
+   When using custom assets with webpack-encore, you need to update the admin build in the public/build/admin directory: :doc:`build-admin-frontend`
+
```
alexander-schranz: Now I understand your request. This is not correct that this is needed and is more a hint the documentation here was not followed or you did run into any cache issues. If you run into that problem that the admin and website build is conflicting, check the changes about the public output directories so that `public/build` is correctly changed to `public/build/website` everywhere it is documented here.

That way the `public/build/website` is for website and `public/build/admin` is for admin and so the admin build is not required as they have both there own.

If you accidently did a build before doing this changes a `git checkout public/build/admin` should restore your files.

cookbook/webpack-encore.rst
```diff
@@ -140,6 +140,10 @@ and run the following command:
 
 Now you should be able to see the outcome in the browser.
 
+.. note::
+
+   When using custom assets with webpack-encore, you need to update the admin build in the public/build/admin directory: :doc:`build-admin-frontend`
```
alexander-schranz:
```suggestion
   If website assets were accidentally build before changing the `public/build` to `public/website/build`
   it could happen that all admin assets were removed. Use `git checkout public/build/admin` or the `bin/console sulu:admin:update-build` command to restore the admin build files: :doc:`build-admin-frontend`
```

alexander-schranz (changes requested):

cookbook/webpack-encore.rst
alexander-schranz: So I would change the text to something like this. Think it is a good hint to add still a note here, for everybody landing into the same situation then you did.

cookbook/webpack-encore.rst
oprichnik: Okay, i see, i remember i forgot one replacement when refactoring from /build/ to /build/website, so this is why my admin build was not working.

Thanks

oprichnik: Just something about git checkout, are the admin build files supposed to be tracked with GIT ?

alexander-schranz: @oprichnik by default our build files are tracked on git. I personally like it this way as you can checkout a project fastly without have the need to rebuild all things until you can work with the project. Some don't like it because they mostly have merge conflicts in them and like it to keep building part of the CI/CD jobs in deployment.

So at the end do what works best for your team, both has its advantages and disadvantages.

alexander-schranz:
@oprichnik thx for report and providing a pull request!

alexander-schranz (approved):

PR #708, opened 2022-04-25
Update the ghost_script configuration for pdf preview

burki:
#### What's in this PR?

The missing path is appended to ghost_script configuration.
In addition, a typo has been fixed.

#### Why?

The pdf-preview-images recipe is lacking the path


niklasnatter (approved):
Thank you!

PR #722, opened 2022-07-11
Fix two factor package name

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets | fixes
| License | MIT

#### What's in this PR?

Fix two factor package name.

#### Why?

False package name in docs.


alexander-schranz:
fixed by #727

PR #723, opened 2022-07-12
Add immutable cache header to nginx

alexander-schranz:
| Q | A
| --- | ---
| Fixed tickets |
| License | MIT

#### What's in this PR?

Add immutable cache header to nginx.

#### Why?

https://github.com/sulu/sulu-docs/pull/700

### DIFF
--- a/cookbook/web-server/nginx.rst
+++ b/cookbook/web-server/nginx.rst
@@ -32,7 +32,7 @@ The Nginx configuration could look something like.
           access_log off;
           expires 1y;
           add_header Pragma public;
-          add_header Cache-Control "public";
+          add_header Cache-Control "public, immutable";
       }
 
       # pass the PHP scripts to FastCGI server from upstream phpfcgi


cookbook/web-server/nginx.rst
```diff
@@ -32,7 +32,7 @@ The Nginx configuration could look something like.
           access_log off;
           expires 1y;
           add_header Pragma public;
-          add_header Cache-Control "public";
+          add_header Cache-Control "public, immutable";
```
luca-rath, line 35: @alexander-schranz Does this rule also apply for the images placed e.g. in `public/website/images/logo.png`? Because this might change and should not use the `immutable` flag i guess

alexander-schranz: They should also be versionated (when replaced via `?v=2`, ...)  as they already have a cachelifetime from 1year.

luca-rath: okay then

luca-rath (approved):

PR #728, opened 2022-07-18
improve indentation

rogamoore:
| Q | A
| --- | ---
| Fixed tickets | -
| Related PRs | -
| License | MIT

#### What's in this PR?

Improve indentation - otherwise it looks like `two_factor` is a child of `logout`

#### Why?

Which problem does the PR fix?

### DIFF
--- a/bundles/security/index.rst
+++ b/bundles/security/index.rst
@@ -128,7 +128,7 @@ admin firewall. This is configured in the ``config/packages/security.yaml``:
         access_control:
             # ...
             - { path: ^/admin/login$, roles: PUBLIC_ACCESS }
-   +         - { path: ^/admin/2fa, role: PUBLIC_ACCESS }
+   +        - { path: ^/admin/2fa, role: PUBLIC_ACCESS }
             # ...
         firewalls:
             # ...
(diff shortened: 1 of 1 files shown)


bundles/security/index.rst
```diff
@@ -137,13 +137,13 @@ admin firewall. This is configured in the ``config/packages/security.yaml``:
...
                 logout:
                     path: sulu_admin.logout
-   +             two_factor:
-   +                 prepare_on_login: true
-   +                 prepare_on_access_denied: true
-   +                 check_path: 2fa_login_check_admin
-   +                 authentication_required_handler: sulu_security.two_factor_authentication_required_handler
-   +                 success_handler: sulu_security.two_factor_authentication_success_handler
-   +                 failure_handler: sulu_security.two_factor_authentication_failure_handler
+   +            two_factor:
```
alexander-schranz, line 140: The additional space here is expected because it shows a `git diff` if you copy the lines into your file you and remove the `+` it will have the correct indention. else it will have a missing space.

alexander-schranz:
there was already open #722 but also fine to merge this ;)

bundles/security/index.rst
rogamoore: Hm, the sphinx diff rendering seems to be a bit confusing then, at least for me :)

Diff on your PR in Github (https://github.com/sulu/skeleton/commit/1572f8aefccb6686cb6a3dbcf5dd466ff1788def?diff=unified):
image

Diff in the Sulu Docs (https://docs.sulu.io/en/latest/bundles/security/index.html?highlight=2fa#two-factor-authentication):
image

Maybe there is some config that can be changed to show it aligned?

alexander-schranz: Sadly the copy & paste view in github is for most diff views completely broken and copy false parts. Atleast copy & paste in sphiny renderer works like expected without any problems here. It just a little bit confusing with indention when looking at the code but I would currently keep it this way to make it easier to copy it instead of go on how it looks.

rogamoore: okay, no problem :)

PR #742, opened 2022-10-10
Add url to sulu_content_load example

niklasnatter:
I think it is not perfectly intuitive how to get the URL of the loaded page. I hope that adding it to the example makes it easier to find out for people.

### DIFF
--- a/reference/twig-extensions/functions/_page_structure.inc
+++ b/reference/twig-extensions/functions/_page_structure.inc
@@ -8,7 +8,6 @@
     - **changer**: User ID of changer
     - **created**: Date of creation
     - **creator**: User ID of creator
-    - **localizations**: Localizations including URLs to other languages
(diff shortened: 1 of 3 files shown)


reference/twig-extensions/functions/_page_structure.inc
```diff
@@ -8,7 +8,6 @@
     - **changer**: User ID of changer
     - **created**: Date of creation
     - **creator**: User ID of creator
-    - **localizations**: Localizations including URLs to other languages
```
niklasnatter, line 11: `localizations` was never part of the `sulu_content_load` response. it is added in the [`ParameterResolver`](https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php) which is only used by the `WebsiteController`, but not the `sulu_content_load` extension.

PR #744, opened 2022-10-25
Add cookbook entry for running Sulu in a multi-server setup

niklasnatter:
It is necessary to configure a few different things to run sulu in a multi-server environment, so we should add a cookbook entry for that 🙂

### DIFF
--- a/cookbook/multi-server-setup.rst
+++ b/cookbook/multi-server-setup.rst
@@ -13,12 +13,36 @@ Media files that are uploaded in the media section of the Sulu administration in
 current server by default. If your application runs on multiple servers, you need to :doc:`configure an external storage adapter<./external-media-storage>`
 for your media files.
 
+.. code-block:: yaml
+
+    # config/packages/sulu_media.yaml
+    sulu_media:
+        storage: s3
+        storages:
+            s3:
+                key: 'your aws s3 key'
+                secret: 'your aws s3 secret'
+                bucket_name: 'your aws s3 bucket name'
+                path_prefix: 'optional path prefix'
+                region: 'eu-west-1'
+
 
 HTTP Cache
 ----------
 
 Sulu utilizes the `FOSHttpCacheBundle`_ to improve response times by caching rendered pages. This cache is automatically invalidated
 when the content of a page is changed. To prevent different cache entries on different servers, you need to :doc:`configure Varnish as a centralized caching proxy<./caching-with-varnish>`.
+
+.. code-block:: yaml
+
+    # config/packages/sulu_http_cache.yaml
+    sulu_http_cache:
+        proxy_client:
+            symfony:
+                enabled: false
+            varnish:
+                enabled: true
+                servers: [ '192.168.0.10:80' ]
 
 
 Application Cache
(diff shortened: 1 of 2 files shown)


cookbook/multi-server-setup.rst
```diff
@@ -0,0 +1,67 @@
...
+If your application runs on multiple servers, you need to `configure a centralized caching adapter`_ like ``redis`` for your ``app`` cache.
+Additionally, if the application runs in different directories on different servers, you need to set a static ``prefix_seed``:
+
+.. code-block:: yaml
+
+    # config/packages/cache.yaml
+    framework:
+        cache:
+            app: cache.adapter.redis
+            prefix_seed: your_vendor_name/app_name
```
alexander-schranz, line 65: would add a link to the `prefix_seed` docs 

https://symfony.com/doc/5.4/reference/configuration/framework.html#prefix-seed

cookbook/multi-server-setup.rst
```diff
@@ -0,0 +1,67 @@
...
+            default_redis_provider: "redis://127.0.0.1:6379"
+
+
+Search Index
+------------
+
+Sulu uses the `MassiveSearchBundle`_ for its search functionality on the website and in the administration interface.
+By default the bundle creates an optimized search index in the current server. To prevent outdated search results,
```
alexander-schranz:
```suggestion
By default the bundle creates an optimized search index on the current server. To prevent outdated search results,
```

Not sure?

cookbook/multi-server-setup.rst
```diff
@@ -0,0 +1,67 @@
...
+
+
+Media Storage
+-------------
+
+Media files that are uploaded in the media section of the Sulu administration interface are stored in the filesystem of the
+current server by default. If your application runs on multiple servers, you need to :doc:`configure an external storage adapter<./external-media-storage>`
+for your media files.
+
+
```
alexander-schranz: Would also add example S3 configuration here.

cookbook/multi-server-setup.rst
```diff
@@ -0,0 +1,67 @@
...
+HTTP Cache
+----------
+
+Sulu utilizes the `FOSHttpCacheBundle`_ to improve response times by caching rendered pages. This cache is automatically invalidated
+when the content of a page is changed. To prevent different cache entries on different servers, you need to :doc:`configure Varnish as a centralized caching proxy<./caching-with-varnish>`.
+
+
```
alexander-schranz: Would also an example `sulu_http_cache` varnish configuration here.

cookbook/multi-server-setup.rst
```diff
@@ -0,0 +1,67 @@
...
+            default_redis_provider: "redis://127.0.0.1:6379"
+
+
+Search Index
+------------
+
+Sulu uses the `MassiveSearchBundle`_ for its search functionality on the website and in the administration interface.
+By default the bundle creates an optimized search index in the current server. To prevent outdated search results,
+you need to `configure Elasticsearch as a centralized search adapter`_.
+
```
alexander-schranz: Would also add example massive_search config here.

cookbook/multi-server-setup.rst
```diff
@@ -0,0 +1,67 @@
...
+Additionally, if the application runs in different directories on different servers, you need to set a static ``prefix_seed``:
+
+.. code-block:: yaml
+
+    # config/packages/cache.yaml
+    framework:
+        cache:
+            app: cache.adapter.redis
+            prefix_seed: your_vendor_name/app_name
+            default_redis_provider: "redis://127.0.0.1:6379"
```
alexander-schranz: would use a `ENV` here like this: https://github.com/sulu/sulu-demo/blob/80b6d2d31412b15e9dbefa0dd7182497f8539547/config/packages/cache.yaml#L15

https://github.com/sulu/sulu-demo/blob/80b6d2d31412b15e9dbefa0dd7182497f8539547/config/packages/cache.yaml#L1-L2

cookbook/multi-server-setup.rst
```diff
@@ -0,0 +1,67 @@
...
+
+By default, Symfony stores active sessions in the filesystem of the current server. To prevent random logouts between requests,
+you need to manage your sessions in a centralized storage that is accessed by all your servers. Have a look at the
+`Store Sessions in a Database`_ section of the Symfony documentation to find out how to store sessions in a database like Redis or MySQL.
+Alternatively, you can directly set a centralizes ``session.save_handler`` in your ``php.ini``:
+
+.. code-block::
+
+    session.save_handler = redis
+    session.save_path = "tcp://127.0.0.1:6379"
```
alexander-schranz: maybe as example for multi server use a external IP private IP: `192.168.0.10`

cookbook/multi-server-setup.rst
niklasnatter: I am not sure about that, I would prefer to keep the example as simple as possible. Otherwise, some people might think they are required to use environment variables and get confused how to configure them. 
If somebody wants to use environment variables, he probably knows that he can just replace the hardcoded value 🙂

alexander-schranz (approved):

PR #748, opened 2022-10-27
Added explanation how sulu-link tag is processed

stollr:
| Q | A
| --- | ---
| Fixed tickets | -
| Related PRs | -
| License | MIT

#### What's in this PR?

Added explanation how sulu-link tag is processed to its documentation.

#### Why?

I have asked myself (and sulu`s slack channel) why there isn't a twig function to turn UUIDs of pages into a URL. The explanation of the process (credits to @niklasnatter) helped me to understand why a twig function would be an issue.


niklasnatter:
Thanks for the improvement! 🙂

PR #756, opened 2022-12-01
Fix default value for mode parameter of route content type

niklasnatter:


luca-rath:
@niklasnatter Thought we tried to fix docs in the lowest possible version, even if those versions are not supported anymore?

niklasnatter:
Maybe I missed an update, but as far as I remember, we only adjusted maintained versions 🙂

luca-rath:
Maybe you're right, just had that in my mind ...

PR #766, opened 2023-03-09
Added basic documentation about the tools provided by the TestBundle

stollr:
| Q | A
| --- | ---
| Fixed tickets | -
| Related PRs | -
| License | MIT

#### What's in this PR?

Explanation how to define the Sulu context in integration and functional tests.

#### Why?

This information was missing before.

### DIFF
--- a/bundles/test.rst
+++ b/bundles/test.rst
@@ -1,0 +1,143 @@
+TestBundle
+==========
+
+Writing automatic tests for a Sulu project will be similar to tests in a usual
+Symfony project. Therefore you should have a look at their `test documentation`_.
+
+
+Sulu's Kernel Context
+---------------------
+
+Sulu adds an additional layer to the service container of Symfony. This is
+the kernel context. It separates the *admin* area from the *website* area.
+
+When you start writing integration or functional tests you have to keep in mind
+that some services are only available in either context.
+
+Integration Tests
+-----------------
+
+In integration tests you may have depencencies on other services. If a service
+is only available in the website context you will get issues, because the
+kernel is in admin context per default.
+
+Therefore Sulu provides an extended `KernelTestCase`_ to specify the kernel
+context.
+
+.. code-block:: php
+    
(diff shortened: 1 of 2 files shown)


bundles/test.rst
```diff
@@ -0,0 +1,94 @@
...
+    use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
+    use Sulu\Component\HttpKernel\SuluKernel;
+    
+    class RegistrationControllerTest extends SuluTestCase
+    {
+        public function testIndexAction(): void
+        {
+            $client = static::createClient([
+                'sulu.context' => SuluKernel::CONTEXT_WEBSITE,
+            ]);
```
alexander-schranz: Thank you for working on this 👍 :

There is a short helper function for this `static::createWebsiteClient()`: https://github.com/sulu/sulu/blob/0b073109244bfa99def7fb63afdc0c62e9b66980/src/Sulu/Bundle/TestBundle/Testing/CreateClientTrait.php#L50

Think it would also be create that we document the other 2 methods in there and that there is also helper function to get a user without manually creating one: `getTestUser` function.

Which is active only on test environment via: https://github.com/sulu/skeleton/blob/c8a9dcc4e03e627924fb28495e3fe0ac98906828/config/packages/security.yaml#L116-L117

stollr: I have updated the PR so that the helper function is shown.

Regarding your second item, I have documented the `getTestUser` function in combination with Symfony's new [`$client->loginUser($testUser)`](https://symfony.com/doc/current/testing.html#logging-in-users-authentication) method. Because I think it is simpler as the `createAuthenticated{Website}Client` methods.

Additionally I have added a further paragraph about the database purging helper.

Prokyonn (approved):
Looks good to me 👍 

@alexander-schranz do you have anything to add?

alexander-schranz (approved):

alexander-schranz:
@stollr Thank you!

PR #769, opened 2023-03-17
Add controller.service_arguments tag

remdan:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

https://symfony.com/doc/current/controller/service.html

If your controllers don't extend the AbstractController class, you must explicitly mark your controller services as public. Alternatively, you can apply the controller.service_arguments tag to your controller services.

#### Why?

If you prefer, you can use the #[AsController] PHP attribute to automatically apply the controller.service_arguments tag to your controller services...


alexander-schranz:
@remdan Tank you!

PR #798, opened 2024-05-07
Simplify global blocks example to ref

alexander-schranz:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Simplify global blocks example to ref.

#### Why?

The docs should concentrate on the global blocks example the additional types just confuses to focus the developer reading the docs to the important part. So moving the mixed part into a comment.

### DIFF
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -382,7 +382,7 @@ from within a block.
 
 .. note::
 
-    Because the ``&`` character needs to be escaped inside of XML files, you have to use ``AND`` instead of ``&&`` 
+    Because the ``&`` character needs to be escaped inside of XML files, you have to use ``AND`` instead of ``&&``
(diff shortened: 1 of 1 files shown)


book/templates.rst
```diff
@@ -942,23 +942,14 @@ node and the ``ref`` attribute:
...
-                    </property>
-                </properties>
-            </type>
         </types>
     </block>
 
+.. note::
+
+    Mixing global and local block types is supported but it is recommended to keep your block type names unique,
+    to avoid confusion and change to global blocks in future easier without data migration.
```
Prokyonn:
```suggestion
    Mixing global and local block types is supported, but it is recommended to keep your block type names unique
    to avoid confusion. This approach also simplifies the transition to global blocks in the future, eliminating the need
    for data migrations.
```

Prokyonn (approved):

PR #802, opened 2024-05-22
update: upgrading node version for admin build

lubomirfiala:
Version of node container recommended by documentation can no longer be used with sulu 2.6..

### DIFF
--- a/cookbook/build-admin-frontend.rst
+++ b/cookbook/build-admin-frontend.rst
@@ -31,7 +31,7 @@ Solution 2: Build manually with docker
 
 .. code-block:: bash
 
-    docker run --rm --interactive --tty --volume ${PWD}:/var/project node:14.16.0 /bin/bash
+    docker run --rm --interactive --tty --volume ${PWD}:/var/project node:20-alpine3.18 /bin/sh
 
     # for completion: using another node version is possible by adjusting the tag of the node image
     # docker run --rm --interactive --tty --volume ${PWD}:/var/project node:12.21.0 /bin/bash


cookbook/build-admin-frontend.rst
```diff
@@ -31,7 +31,7 @@ Solution 2: Build manually with docker
 
 .. code-block:: bash
 
-    docker run --rm --interactive --tty --volume ${PWD}:/var/project node:14.16.0 /bin/bash
+    docker run --rm --interactive --tty --volume ${PWD}:/var/project node:20-alpine3.18 /bin/sh
 
     # for completion: using another node version is possible by adjusting the tag of the node image
     # docker run --rm --interactive --tty --volume ${PWD}:/var/project node:12.21.0 /bin/bash
```
alexander-schranz, line 37: we should also change this line e.g. using `18.19.0`

cookbook/build-admin-frontend.rst
```diff
@@ -31,7 +31,7 @@ Solution 2: Build manually with docker
 
 .. code-block:: bash
 
-    docker run --rm --interactive --tty --volume ${PWD}:/var/project node:14.16.0 /bin/bash
+    docker run --rm --interactive --tty --volume ${PWD}:/var/project node:20-alpine3.18 /bin/sh
```
alexander-schranz, line 34: I would not add `-alpine` as they might sometimes make problems so I would keep with the default image and the `/bin/bash`

alexander-schranz (changes requested):
Thx for the merge request, some small change requests.

alexander-schranz:
node versions where upgraded in https://github.com/sulu/sulu-docs/pull/897.
Still thx for the PR

PR #809, opened 2024-06-25
Fix documentation for excluding templates in Sulu 2.x

### LINKED ISSUE #808: Removal of the caution regarding template visibility in the template selection dropdown
According to the bug report opened here: https://github.com/sulu/sulu/issues/7476

The observed behavior is the expected one.
This is not a bug in Sulu, but rather an inconsistency in the documentation regarding the "Caution" section of the "Choosing the Template of a Page" title in the templates.rst file.

Since Sulu 2.x to exclude specific templates, you must configure as excluded in the webspace.xml file using the excluded-templates node.

mcoiffier:
| Q | A
| --- | ---
| Fixed tickets | fixes [#888](https://github.com/sulu/sulu-docs/issues/808)
| License | MIT

#### What's in this PR?
Removal of the caution regarding the visibility of templates in the dropdown list.

#### Why?
Because since version 2.x of Sulu, this is no longer the case. You need to use the `excluded-templates` node in the webspace file.

### DIFF
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -40,6 +40,11 @@ The template of a page can be selected in the admin interface:
 
 .. figure:: ../img/templates-selection.png
 
+.. note::
+
+    If you don't want a template to appear in this list, use the ``excluded-templates`` node in the webspace configuration file.
+    Have a closer look at `Setup a Webspace <webspaces.html#urls>`_ for more details.
+
 The name displayed in the dropdown is configured in the ``<meta>`` section of
 the XML:
 
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -40,12 +40,6 @@ The template of a page can be selected in the admin interface:
 
 .. figure:: ../img/templates-selection.png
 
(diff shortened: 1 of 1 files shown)


book/templates.rst
```diff
@@ -40,12 +40,6 @@ The template of a page can be selected in the admin interface:
 
 .. figure:: ../img/templates-selection.png
 
-.. Caution::
-
-    A template is shown in the dropdown only if both the XML and the Twig file
-    exist! If you can't see your template, double-check the directories
-    ``config/templates/pages`` and ``templates/pages``.
-
```
alexander-schranz: Should we add a info or hint for the excluded part in the webspace?

mcoiffier: So, what do you think about this commit 3c5641b34847a4579b117e741416923c9f9052af?

alexander-schranz (approved):

alexander-schranz:
@mcoiffier Thank you!

PR #823, opened 2024-11-28
Fix analytics template path

fnagel:
See https://github.com/sulu/sulu/discussions/7692

### DIFF
--- a/cookbook/manage-analytics.rst
+++ b/cookbook/manage-analytics.rst
@@ -39,10 +39,10 @@ You are able to override the analytics template with the
 
 There are three relevant templates folders:
 
-* ``SuluWebsiteBundle:Analytics/type/google``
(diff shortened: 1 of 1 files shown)


alexander-schranz:
Thank you can we use the more modern syntax with `/` instead of `:` and can you rebase and target the 2.5 branch?

fnagel:
Changed the code, but unsure if I did right with the branch as there is that "This branch is out-of-date with the base branch" note form Github.

fnagel:
Maybe I need to close this one and create a new one like this: https://github.com/sulu/sulu-docs/compare/2.5...fnagel:sulu-docs:patch-1?expand=1

alexander-schranz:
@fnagel looks good. Thank you!

alexander-schranz (approved):

PR #826, opened 2024-12-11
Adjust webspace naming

Prokyonn:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

In the new Sulu Skeleton the webspace name was changes from `example` to `website`.

### DIFF
--- a/book/webspaces.rst
+++ b/book/webspaces.rst
@@ -91,7 +91,7 @@ following paragraphs.
                     </environment>
                     <environment type="dev">
                         <urls>
-                            <url language="en">website.lo</url>
+                            <url language="en">example.localhost</url>
                         </urls>
                     </environment>
                 </environments>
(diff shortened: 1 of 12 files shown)


book/webspaces.rst
```diff
@@ -270,7 +270,7 @@ localization. The following fragment shows again how to this:
 
 .. code-block:: xml
 
-    <url language="de" country="at">www.example.org</url>
+    <url language="de" country="at">www.website.org</url>
```
alexander-schranz: would still use `example.org` in the official examples as it is official domain special for this kind of things :)

alexander-schranz (changes requested):
We should in example configuration still the `example.org` domain: 

`website.com`, `website.org` -> `example.org`

As that is official standard Domain for examples and can not be Hijacking by somebody in the future.

alexander-schranz (approved):

PR #829, opened 2025-01-07
Adding documentation for FrankenPHP

mamazu:
| Q | A
| --- | ---
| Fixed tickets | -
| Related PRs | -
| License | MIT

#### What's in this PR?
Explaining how adding frankenphp works.

#### Why?
This is also officially supported by Sulu so we might as well document it.

### DIFF
--- a/cookbook/web-server/franken-php.rst
+++ b/cookbook/web-server/franken-php.rst
@@ -1,0 +1,63 @@
+Running Franken PHP
+===================
+
+To add FrankenPHP to an existing project you can follow these steps: `FrankenPHP Documentation`_.
+
+However after running through those steps you need to modify the setup a slight bit:
+1. Installing the correct php extensions
+2. Adding a database to the project
+3. Removing Sulu's default configuration
+
+The modifications
+-----------------
+.. note::
+    Beware these instructions might get out of date if FrankenPHP Documentation changes.
+
+After installing you should have a `Dockerfile` which is the base for all containers.
+
+There should be a section under like this:
+.. code-block:: dockerfile
+    # Base FrankenPHP image
+    ...
+    RUN set -eux; \
+        install-php-extensions \
+            @composer \
+            apcu \
+            intl \
+            opcache \
+            zip \
+        ;
+
+This is the list of installed extensions. For Sulu you also need: `gd` and `pdo_mysql` or `pdo_postgres`.
+
+Secondly we also need to add a database. We can just do that by adding it to the yaml configuration file `compose.yaml`.
+
+Here is an example configuration for mysql:
+
+.. code-block:: yaml
+       services:
+          # ....
+          database:
+            # arm compatible mysql docker image
+            image: mysql/mysql-server:${MYSQL_VERSION:-8.0} # arm and x86/x64 compatible mysql image
+            command: [ "--max_connections=1000", "--character-set-server=utf8mb4", "--collation-server=utf8mb4_general_ci" ]
+            environment:
+              MYSQL_DATABASE: ${MYSQL_DATABASE:-sulu}
+              # You should definitely change the password in production
+              MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-ChangeMe}
+              MYSQL_ROOT_HOST: '%'
+            volumes:
+              - db-data:/var/lib/mysql
+              # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
+              # - ./docker/db/data:/var/lib/mysql:rw
+
+After setting up the container you might also want to update the `DATABASE_URL` variable in the php service (by default it assumes postgres). At least note that FrankenPHP and the default Sulu mysql container have different default credentials.
+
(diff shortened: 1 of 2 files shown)


cookbook/web-server/franken-php.rst
```diff
@@ -0,0 +1,63 @@
+Running Franken PHP
```
alexander-schranz: would call this FrankenPHP via Docker

as you can use also frankenphp standalone binary which is different to use. Same for the file name.

cookbook/web-server/franken-php.rst
```diff
@@ -0,0 +1,63 @@
...
+              # - ./docker/db/data:/var/lib/mysql:rw
+
+After setting up the container you might also want to update the `DATABASE_URL` variable in the php service (by default it assumes postgres). At least note that FrankenPHP and the default Sulu mysql container have different default credentials.
+
+If you have some other modifications in your set up be sure to copy all changes of the `docker-compose*` files to the `compose.yaml` and `compose.override.yaml`.
+
+After all this, the Sulu default configuration is not used anymore and can be removed:
+.. code-block:: bash
+    rm docker-compose.yaml
+    rm docker-compose.override.yaml
```
alexander-schranz: Not understanding this parts sulu use `compose.yaml` files like Symfony and Dunglas Symfony Docker. The docker-compose.yaml are not in use by any of them.

alexander-schranz (changes requested):

cookbook/web-server/franken-php.rst
```diff
@@ -0,0 +1,63 @@
...
+The modifications
+-----------------
+.. note::
+    Beware these instructions might get out of date if FrankenPHP Documentation changes.
+
+After installing you should have a `Dockerfile` which is the base for all containers.
+
+There should be a section under like this:
+.. code-block:: dockerfile
+    # Base FrankenPHP image
```
alexander-schranz:
```suggestion
There should be a section under like this:

.. code-block:: dockerfile

    # Base FrankenPHP image
```

cookbook/web-server/franken-php.rst
```diff
@@ -0,0 +1,63 @@
...
+        ;
+
+This is the list of installed extensions. For Sulu you also need: `gd` and `pdo_mysql` or `pdo_postgres`.
+
+Secondly we also need to add a database. We can just do that by adding it to the yaml configuration file `compose.yaml`.
+
+Here is an example configuration for mysql:
+
+.. code-block:: yaml
+       services:
```
alexander-schranz:
```suggestion
.. code-block:: yaml

       services:
```

cookbook/web-server/franken-php.rst
```diff
@@ -0,0 +1,63 @@
...
+              # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
+              # - ./docker/db/data:/var/lib/mysql:rw
+
+After setting up the container you might also want to update the `DATABASE_URL` variable in the php service (by default it assumes postgres). At least note that FrankenPHP and the default Sulu mysql container have different default credentials.
+
+If you have some other modifications in your set up be sure to copy all changes of the `docker-compose*` files to the `compose.yaml` and `compose.override.yaml`.
+
+After all this, the Sulu default configuration is not used anymore and can be removed:
+.. code-block:: bash
+    rm docker-compose.yaml
```
alexander-schranz:
```suggestion
After all this, the Sulu default configuration is not used anymore and can be removed:

.. code-block:: bash

    rm docker-compose.yaml
```

cookbook/web-server/franken-php.rst
```diff
@@ -0,0 +1,63 @@
...
+
+However after running through those steps you need to modify the setup a slight bit:
+1. Installing the correct php extensions
+2. Adding a database to the project
+3. Removing Sulu's default configuration
+
+The modifications
+-----------------
+.. note::
+    Beware these instructions might get out of date if FrankenPHP Documentation changes.
```
alexander-schranz:
```suggestion
-----------------

.. note::

    Beware these instructions might get out of date if FrankenPHP Documentation changes.
```

alexander-schranz (changes requested):

cookbook/web-server/franken-php.rst
mamazu: Yes, that's why I would recommend to delete them. That way we don't confuse the user of which of the compose file is now being used.

cookbook/web-server/franken-php.rst
mamazu: Yes, that's a good idea. Did that.

cookbook/web-server/franken-php.rst
alexander-schranz: I would remove that sentence about docker+compose, we should not consider here that somebody uses may older version of something. This docs is more used for new setup projects which are uptodate. Think else this confuses more.

mamazu: The point should be that whatever you had in your old Sulu `docker-compose.yaml` had will no longer be used when running the `compose.yaml`. That's why you need to merge any changes from the Sulu config into the FrankenPHP config. And if you remove the old file you get conflicts when Sulu changes something in the default config.

PR #838, opened 2025-04-24
Provide documentation also as composer package

alexander-schranz:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Provide documentation also as composer package.

#### Why?

Lets discuss this, I think in the raise of AI tools it make sense that somebody maybe can install the documentation via composer to give it as a context to AI tools.

The sulu docs still will not be versionized and it will only be installable as `@dev`.

```bash
composer require sulu/sulu-docs:"2.6.*@dev" --dev
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ _build/
 *.iml
 .idea
 .DS_STORE
+
+# Composer
+/vendor/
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@alexander-schranz yeah thats really good - would just do IT!

PR #842, opened 2025-08-06
Add hint about cleanup temp files

alexander-schranz:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Add hint about cleanup temp files.

#### Why?

I discuss with Yanick from Contao CMS this issue, and they say its a common problem and if imagick is used they recommend a cronjob to cleanup old files.

### DIFF
--- a/bundles/media/imagine-adapter.rst
+++ b/bundles/media/imagine-adapter.rst
@@ -59,7 +59,7 @@ To install it use following commands:
 
 .. note::
 
-    It is very common that the `imagick` extension my not remove failed temporary files.
+    It is very common that the `imagick` extension may not remove failed temporary files.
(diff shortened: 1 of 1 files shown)


bundles/media/imagine-adapter.rst
```diff
@@ -57,6 +57,16 @@ To install it use following commands:
     apt-get install libmagickwand-dev
     apt-get install php8.3-imagick
 
+.. note::
+
+    It is very common that the `imagick` extension my not remove failed temporary files.
```
Prokyonn:
```suggestion
    It is very common that the `imagick` extension may not remove failed temporary files.
```

Prokyonn (approved):

PR #847, opened 2025-09-02
Add documentation about image formats for multi server setup

alexander-schranz:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Add documentation about image formats for multi server setup.

#### Why?

Its a common issue that media storage and image formats are the same.


cookbook/multi-server-setup.rst
```diff
@@ -26,6 +26,30 @@ for your media files.
...
                 region: 'eu-west-1'
 
+Image Formats
+-------------
+
+In the Media Storage only original files are stored. The image formats are generated by Sulu the first time they got
+requested. That mechanic (reverse proxy cache) storages like S3, Blob Storage, ... do not support.
+So for every server the image will be regenerated once. Aslong as no media are removed that might not be
+a problem. You can per server configure ``bin/console sulu:media:format:cache:cleanup`` as a cronjob to make sure
+that a image format is also removed on all servers.
```
Prokyonn:
```suggestion
In the Media Storage, only original files are stored. The image formats are generated by Sulu the first time they are
requested. That mechanism (reverse proxy cache) is not supported by storages like S3, Blob Storage, etc. So for
every server, the image will be regenerated once. As long as no media are removed, that might not be a problem.
You can configure ``bin/console sulu:media:format:cache:cleanup`` as a cron job per server to ensure that an image
format is also removed on all servers.
```

cookbook/multi-server-setup.rst
```diff
@@ -26,6 +26,30 @@ for your media files.
...
+
+Still in a kubernetes / container setups where you might loose the ``public/uploads/media`` everytime a container restart,
+you need to find different ways to cache images.
+You can make the ``public/uploads/media`` directory a mounted volume which is shared between all containers (might not be good for performance).
+Better use some kind of CDN (cloudflare, fastly, ...) or a reverse proxy cache via nginx to cache generate images formats.
+
+If you configure a CDN / reverse proxy correctly you not require the cronjob and can disable the image format filesystem cache completely via:
```
Prokyonn:
```suggestion
Still, in Kubernetes/container setups where you might lose the ``public/uploads/media`` directory every time a container restarts,
you need to find different ways to cache images.
You can make the ``public/uploads/media`` directory a mounted volume shared between all containers (which might not be good for performance).
It is better to use some kind of CDN (Cloudflare, Fastly, etc.) or a reverse proxy cache via Nginx to cache generated image formats.

If you configure a CDN/reverse proxy correctly, you do not require the cron job and can disable the image format filesystem cache completely via:
```

cookbook/multi-server-setup.rst
```diff
@@ -26,6 +26,30 @@ for your media files.
...
+
+.. code-block:: yaml
+
+    # config/packages/sulu_media.yaml
+    sulu_media:
+        format_cache:
+            save_image: false
+
+Never disable the image format cache if you do not use any CDN or reverse proxy cache which caches files a very long time.
```
Prokyonn:
```suggestion
Never disable the image format cache if you do not use a CDN or reverse proxy cache that caches files for a very long time.
```

PR #858, opened 2025-12-02
Update documentation for the RouteBundle

Prokyonn:
| Q | A
| --- | ---
| Related PRs | https://github.com/sulu/sulu/pull/8428
| License | MIT

#### What's in this PR?

Adjusts documentation for the new RouteBundle of Sulu 3.0

#### Why?

Current version is outdated.


alexander-schranz (approved):

bundles/route/custom_route_generator.rst
```diff
@@ -1,67 +0,0 @@
-Custom Route-Generator
-======================
```
alexander-schranz, line 2: create a `:orphan:` here and link to the route here. So switching from `2.x` -> `3.x` give a hint where to find new docs.

reference/property-types/resource_locator.rst
```diff
@@ -1,61 +0,0 @@
-Resource locator
-================
```
alexander-schranz, line 2: create a `:orphan:` here and link to the route here. So switching from `2.x` -> `3.x` give a hint where to find new docs.

alexander-schranz (changes requested):

bundles/route/custom_route_generator.rst
Prokyonn: Added the two files. Alternatively, we could add redirects instead of leaving them as orphaned files.

alexander-schranz (approved):

bundles/route/custom_route_generator.rst
alexander-schranz: thx, redirects are limited on readthedocs so if we can use orphans we should :)

PR #865, opened 2026-01-07
Change IP address to localhost in getting-started.rst

truckee:
Updated database connection URLs from '127.0.0.1' to 'localhost' for consistency.

| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Use "localhost" instead of "127.0.0.1" for universal (IPV6 or IPV4) URLs.

#### Why?

Systems only recognizing IPV6 will not accept IPV4 URLs.

### DIFF
--- a/book/getting-started.rst
+++ b/book/getting-started.rst
@@ -118,7 +118,7 @@ Here is an example for using Sulu with MySQL:
 
 .. code:: bash
 
-    DATABASE_URL=mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7
+    DATABASE_URL=mysql://db_user:db_password@localhost:3306/db_name?serverVersion=5.7
 
 When you're done with the configuration, populate the database with Sulu's
 default data:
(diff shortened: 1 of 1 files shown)


book/getting-started.rst
```diff
@@ -118,7 +118,7 @@ Here is an example for using Sulu with MySQL:
 
 .. code:: bash
 
-    DATABASE_URL=mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7
+    DATABASE_URL=mysql://db_user:db_password@localhost:3306/db_name?serverVersion=5.7
```
alexander-schranz, line 121: We use Symfony defaults here always, which we stay with if you want to change this from 127.0.0.1 to localhost you need create a Pull Request to Symfony and check with the Symfony / Doctrine Team why they use 127.0.0.1 instead of localhost.

Check: https://github.com/symfony/recipes/blob/db54e87e75fbb63c8ab84def24d7cf4632fd5dcc/doctrine/doctrine-bundle/2.13/manifest.json#L14-L16

book/getting-started.rst
```diff
@@ -153,10 +153,10 @@ will make sure the correct application is loaded.
 
     php -S localhost:8000 -t public/ config/router.php
```
alexander-schranz, line 154: Would more go with `127.0.0.1` here instead of `localhost`.

alexander-schranz (changes requested):

alexander-schranz:
PS: bugfixes in docs need target lowest supported version which is currently `2.6`.

book/getting-started.rst
truckee: A modification is indicated because Windows 11 provides a naked  **hosts** file. That file at a minimum needs to contain:
```
127.0.0.1 localhost
::1 localhost
```
Shall I take this to Symfony or redo my pull request here?

PR #891, opened 2026-02-12
Update property types about articles

martinlagler:
#### What's in this PR?

Update property types for articles for Sulu 3.x

### DIFF
--- a/reference/property-types/article_selection.rst
+++ b/reference/property-types/article_selection.rst
@@ -104,6 +104,6 @@ Twig
                 {{ article.excerptDescription|default(article.article)|raw }}
             </div>
 
-            <div href="{{ sulu_content_path(article.url) }}">Read more</div>
+            <a href="{{ sulu_content_path(article.url) }}">Read more</a>
         </div>
     {% endfor %}
--- a/reference/property-types/article_selection.rst
+++ b/reference/property-types/article_selection.rst
@@ -1,10 +1,10 @@
-Page Selection
-==============
+Article selection
+=================
 
 Description
 -----------
 
-Shows a list with the possibility to add links to other pages managed in Sulu.
+Shows a list with the possibility to add links to other articles managed in Sulu.
 Additionally it populates all the fields defined in the template configuration
 to the HTML template. The content is stored as an array of references.
 
@@ -19,7 +19,7 @@ Parameters
       - Description
     * - properties
       - collection
-      - Defines with which key which property of the linked page should be
+      - Defines with which key which property of the linked article should be
         populated to the HTML template.
     * - item_disabled_condition
       - string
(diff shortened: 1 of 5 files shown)


reference/property-types/article_selection.rst
```diff
@@ -0,0 +1,85 @@
...
+.. _jexl: https://github.com/TomFrost/jexl
+
+Twig
+----
+
+.. code-block:: twig
+
+    {% for article in content.articles %}
+        <a href="{{ sulu_content_path(article.url) }}">
+            <h3>
```
alexander-schranz, line 99: invalid HTML, block tags (`h3`) are not allowed inside inline tags (<`a>`)

```
<div>
    <h3>
    <div>
    <a href="">Read more</a>
</div>
```

reference/property-types/single_page_selection.rst
```diff
@@ -35,27 +41,34 @@ Example
...
+
+            <params>
+                <param name="properties" type="collection">
+                    <param name="title" value="title"/>
+                    <param name="article" value="article"/>
+                    <param name="excerptTitle" value="excerpt.title"/>
+                    <param name="excerptDescription" value="excerpt.description"/>
+                    <param name="uuid" value="object.resource.id"/>
+                </param>
+                <param name="item_disabled_condition" value="title == 'No'"/>
```
alexander-schranz, line 77: provide always a simple example first which shows the basic usecase. And only uncommon things like object.resource and item_disabled_conditoin mapping to a own `Complex Example` under own headline.

As the first example is mostly the one which is copied to projects and it should not contain things which are not very common.

reference/property-types/single_page_selection.rst
```diff
@@ -35,27 +41,34 @@ Example
...
-
-    <sulu-link href="{{ content.link }}">Link Text</sulu-link>
-
-If you need to load additional data of the target page, you can use the
-:doc:`sulu_content_load twig extension<../twig-extensions/functions/sulu_content_load>`:
-
 .. code-block:: twig
 
-    {% set target = sulu_content_load(content.link, {'title': 'title', 'excerptTitle': 'excerpt.title'}) %}
+    <a href="{{ sulu_content_path(content.page.url) }}">
```
alexander-schranz: same as for article we should create valid HTML and keep things like no block tags inside inline tags in mind.

alexander-schranz (changes requested):

reference/property-types/article_selection.rst
```diff
@@ -0,0 +1,109 @@
...
+        <div>
+            <h3>
+                {{ article.excerptTitle|default(article.title) }}
+            </h3>
+
+            <div>
+                {{ article.excerptDescription|default(article.article)|raw }}
+            </div>
+
+            <div href="{{ sulu_content_path(article.url) }}">Read more</div>
```
alexander-schranz:
```suggestion
            <a href="{{ sulu_content_path(article.url) }}">Read more</a>
```

alexander-schranz (approved):

PR #906, opened 2026-04-21
Add article documentation

martinlagler:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Documentation for the ArticleBundle, Groups, group translations, webspace configs.

### DIFF
--- a/bundles/article/index.rst
+++ b/bundles/article/index.rst
@@ -1,0 +1,119 @@
+ArticleBundle
+=============
+
+The ArticleBundle provides a dedicated space for managing content-rich entities such as blog posts,
+news articles, or guides. Unlike pages, which are organized in a hierarchical tree, articles are
+managed as a flat list — making it well-suited for large volumes of content that don't belong to
+a fixed site structure.
+
+Articles support drafting, publishing, versioning, localizations (including shadow and ghost pages),
+SEO and excerpt management, author and publication date handling, as well as search, filter, and
+sort capabilities in the Sulu admin.
+
+Groups
+------
+
+Groups allow you to organize article templates into separate sections in the Sulu admin interface.
+Each group gets its own navigation entry and list view. A group is assigned to a template by adding
+the ``<group>`` tag to the template XML:
+
+.. code-block:: xml
+
+    <?xml version="1.0" ?>
+    <template xmlns="http://schemas.sulu.io/template/template"
+              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">
+
+        <key>blog</key>
+        <group>blog</group>
+
+        <!-- ... -->
+
+    </template>
+
+All templates sharing the same group identifier will appear together under one admin section.
+Templates without a ``<group>`` tag are placed into the ``default`` group.
+
+The group label shown in the admin is resolved via the translation key
+``sulu_admin.template_group.<group_identifier>``. If no translation is found, the group
+identifier is used with its first letter capitalized.
+
+To provide a custom label, add the translation key to your admin translation files:
+
+.. code-block:: js
+
+    // translations/admin.en.json
+    {
+        "sulu_admin.template_group.blog": "Blog",
+        "sulu_admin.template_group.guide": "Guides"
+    }
+
+Webspace Configuration
+----------------------
+
+Articles are global objects — similar to snippets — that can be assigned to one or more
+webspaces. Each article has a main webspace and optionally one or more additional webspaces;
+it is accessible via all of them. The active webspaces per article can be configured
+individually in the article's **Settings** tab in the Sulu admin, or set globally via
+``config/packages/sulu_article.yaml``. Note that ``default_main_webspace`` is required when
+using the article bundle.
+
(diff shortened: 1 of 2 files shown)


bundles/article/index.rst
```diff
@@ -0,0 +1,138 @@
...
+        "sulu_admin.template_group.blog": "Blog",
+        "sulu_admin.template_group.guide": "Guides"
+    }
+
+Webspace Configuration
+----------------------
+
+By default, articles are not linked to a specific webspace. You can configure the default main
+webspace and additional webspaces that articles are published to in
+``config/packages/sulu_article.yaml``.
```
alexander-schranz: > articles are not linked to a specific webspace.

Think this is misleading.

I would maybe more go with something like "Articles are global objects like snippets, .... they can be in one or multiple webspace defined via the settings tab. To avoid duplicate content issue, visiting the article on additional webspace URL Sulu will automatically set a canonical tag to configured main webspace. By default an article main webspace is the first webspace (not sure about this) and is still available in all additional webspaces (not sure abou this).

bundles/article/index.rst
```diff
@@ -0,0 +1,138 @@
...
+    # Different default webspace per locale
+    sulu_article:
+        default_main_webspace:
+            default: my_webspace
+            de: my_webspace_de
+
+The ``default`` key is used as a fallback for locales that have no explicit entry.
+
+Default Additional Webspaces
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
alexander-schranz: would not split this into seperate sections / headlines. The "Webspace Configuration" headline is enough:

Would create one code block show "Main" and "Additional" with single selection:

```yaml
    sulu_article:
        default_main_webspace: my_webspace
        default_additional_webspaces:
            - my_second_webspace
            - my_third_webspace
```

And one per locale:

```yaml
    sulu_article:
        default_main_webspace:
            default: my_webspace
            de: my_webspace_de
        default_additional_webspaces:
            default:
                - my_second_webspace
            de:
                - my_second_webspace_de
```

bundles/article/index.rst
```diff
@@ -0,0 +1,138 @@
...
+
+**Adding an additional webspace:** Insert a new row into
+``ar_article_dimension_content_additional_webspaces`` for every ``ar_article_dimension_contents``
+record where ``customizeWebspaceSettings`` is ``false``. The ``name`` column holds the webspace
+key and ``article_dimension_content_id`` references the dimension content record.
+
+**Removing an additional webspace:** Delete all rows from
+``ar_article_dimension_content_additional_webspaces`` that match the removed webspace key in the
+``name`` column and belong to a dimension content record where ``customizeWebspaceSettings``
+is ``false``.
```
alexander-schranz: Would put this into a `.. note` in the section above, it is related to Webspace Configuration and so should be part of that section not an own section.

alexander-schranz (changes requested):
Some text change suggestions. And I would reduce the headlines as this mostly to the deeplinks which we mostly will send when questions come up:

 - ArticleBundle
     - Webspace Configuration
     - Groups

bundles/article/index.rst
```diff
@@ -0,0 +1,138 @@
...
+The group label shown in the admin is resolved via the translation key
+``sulu_admin.template_group.<group_identifier>``. If no translation is found, the group
+identifier is used with its first letter capitalized.
+
+To provide a custom label, add the translation key to your admin translation files:
+
+.. code-block:: json
+
+    // translations/admin.en.json
+    {
```
alexander-schranz: if use a JS comment maybe better use `js` here or move path to translation file into text above:

```suggestion
.. code-block:: js
```

bundles/article/index.rst
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+Webspace Configuration
+----------------------
+
+Articles are global objects — similar to snippets — that can be assigned to one or more
+webspaces. Each article has a main webspace and optionally one or more additional webspaces;
+it is accessible via all of them. The active webspaces per article can be configured
+individually in the article's **Settings** tab in the Sulu admin, or set globally via
+``config/packages/sulu_article.yaml``. Note that ``default_main_webspace`` is required when
```
martinlagler, line 58: @alexander-schranz Should I maybe put this `Note that ``default_main_webspace`` is required when
using the article bundle.` also in a warning? As it's pretty important.

alexander-schranz: 👍  but its only required when there are multiple webspaces? Else the skeleton is would be broken by default or? https://github.com/sulu/skeleton/blob/3.0/config/packages/sulu_article.yaml

bundles/article/index.rst
```diff
@@ -0,0 +1,119 @@
...
+using the article bundle.
+
+To avoid duplicate content issues, Sulu will automatically set a canonical tag pointing to
+the main webspace URL when an article is accessed via an additional webspace URL.
+
+Both options can be provided as a simple string/list (applies to all locales) or as a
+locale-keyed mapping. The ``default`` key is used as a fallback for locales with no explicit
+entry:
```
alexander-schranz, line 66: would not mention the structure here like `locale-keyed mapping` or `simple string/list (applies to all locales)` as that is clearer by look at the example below. 

Would go with just something like:

> Both options allow to set a seperate value also per locale.

PR #914, opened 2026-05-12
Updated activity docs

martinlagler:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Update activity docs.

### DIFF
--- a/bundles/activity.rst
+++ b/bundles/activity.rst
@@ -16,6 +16,9 @@ The ActivityBundle allows for the following configuration:
         storage:
             adapter: 'doctrine' # can be set to null to not store activities
             persist_payload: false # include payload of event in stored activity
+        objects:
+            activity:
+                model: Sulu\Bundle\ActivityBundle\Domain\Model\Activity # override to use a custom activity entity
 
 Listen for an event
 -------------------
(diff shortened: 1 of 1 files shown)


bundles/activity.rst
```diff
@@ -49,15 +52,18 @@ sends an email when a page with a specific template is created in the applicatio
 
         public function sendPageCreatedMail(PageCreatedEvent $event): void
         {
-            if ('product' === $event->getPageDocument()->getStructureType()) {
+            $dimensionContent = $event->getPage()->getDimensionContents()
+                ->findFirst(fn($key, $dc) => $dc->getLocale() === $event->getResourceLocale());
```
alexander-schranz, line 56: Unsure about this, doesn't this overfetch the collection? And it does not keep dimension in mind, should we add a better way access the correct dimension contents here.

/cc @Prokyonn

bundles/activity.rst
```diff
@@ -16,6 +16,9 @@ The ActivityBundle allows for the following configuration:
         storage:
             adapter: 'doctrine' # can be set to null to not store activities
             persist_payload: false # include payload of event in stored activity
+        objects:
+            activity:
+                model: Sulu\Bundle\ActivityBundle\Domain\Model\Activity # override to use a custom activity entity
```
Prokyonn, line 21: shouldn't this be a custom namespace then?

bundles/activity.rst
Prokyonn: The page should already have eager-loaded the dimensionContents, therefore I think no overfetching here, but still ideally this should use the aggregated dimension (localized+unlocalized) for the given attributes.

bundles/activity.rst
alexander-schranz: For the configruation documentation this should be represent default values.

bundles/activity.rst
alexander-schranz: Maybe we can show how to get a merged dimension here via ContentMerger Service?

Prokyonn: yes that would then either be `ContentAggregator->aggregate($page, $dimensionContentAttributes)` or via the `ContentManager->resolve`. Tbh I don't like the ContentManager, but I think most projects will use it as a facade

PR #917, opened 2026-05-21
Optimize docs for SSO and 2FA

therouv:
| Q | A
| --- | ---
| Fixed tickets | –
| Related PRs | –
| License | MIT

While implementing Single Sign-On (SSO) and Two-Factor Authentication (2FA) in several projects, I noticed a few areas where the current documentation could be improved. Some behavior was not immediately clear from the docs and only became apparent after reviewing the Sule core code or testing with different SSO providers.

In addition, I think the SSO and 2FA sections would benefit from being moved out of the main security documentation into dedicated sub-pages for better structure and readability.

To make the review easier, I split the changes into two separate commits:

* `a02d09c` for moving the pages
* `5eb5652` for the actual documentation improvements


alexander-schranz:
@therouv Can you move the edited block back to the previous file? Splitting docs up has some disadvantages and like Symfony the last years better to have bigger component docs to give a better overall few over all features as splittings things up. Also movement of docs part and changes should never be done in same pull requests as it make reviews focus on false parts.

therouv:
@alexander-schranz Alright, thanks for the heads-up. I’ve merged everything back into the main file, including my changes

PR #920, opened 2026-05-29
Update custom page data provider docs

martinlagler:
| Q | A
| --- | ---
| License | MIT

#### What's in this PR?

Update custom page data provider docs.

#### Why?

To fit Sulu 3.x

### DIFF
--- a/cookbook/custom-page-data-provider.rst
+++ b/cookbook/custom-page-data-provider.rst
@@ -1,15 +1,143 @@
-Filter pages by a property using a custom SmartContent DataProvider
-===================================================================
+Filter pages by a property using a custom Page SmartContent Provider
+======================================================================
 
-The ``smart_content`` property type allows for the configuration of a ``provider`` param, which is used for gathering
-the items that are passed to the twig template. As described in :doc:`/reference/property-types/smart_content`, Sulu
-comes with a few included providers and also allows to implement new ones to load custom entities.
-
-In some cases, it might be useful to register an additional provider, which only returns pages including a property
-with a specific value. This is possible by registering a new instance of the ``PageDataProvider`` included in Sulu
-that uses a custom implementation of the ``QueryBuilder``. In the following example a custom ``AuthorPageDataProvider``
-is implemented, which filters the pages for a specific author:
+The ``smart_content`` property type supports a ``provider`` param that controls which items are
+loaded. In some cases it is useful to register an additional provider that only returns pages
+matching a specific criterion. The following example creates an ``AuthorPagesSmartContentProvider``
+that restricts results to pages written by a specific author.
 
 .. note::
 
-    This documentation does not yet exist for Sulu 3.0. Please feel free to contribute it.
+    The same pattern applies to articles — extend
+    ``Sulu\Article\Infrastructure\Sulu\Content\ArticleSmartContentProvider`` and use
+    ``parent: sulu_article.article_smart_content_provider`` in the service definition.
+
+Example
+-------
+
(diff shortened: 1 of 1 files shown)


cookbook/custom-page-data-provider.rst
```diff
@@ -1,15 +1,102 @@
...
+        protected function addInternalFilters(QueryBuilder $queryBuilder, array $filters, string $alias): void
+        {
+            parent::addInternalFilters($queryBuilder, $filters, $alias);
+
+            $queryBuilder
+                ->leftJoin('filterDimensionContent.navigationContexts', 'navCtx')
+                ->andWhere('navCtx.navigationContext = :navigationContext')
+                ->setParameter('navigationContext', 'footer');
+
+            $this->addInternalSortBys($queryBuilder, $alias);
```
Prokyonn: addInternalSortBys is already called in the parent `PageSmartContentProvider` so we do not have to add it here

cookbook/custom-page-data-provider.rst
```diff
@@ -1,15 +1,102 @@
...
+        {
+            $queryBuilder->addOrderBy($alias . '.lft', 'asc');
+        }
+    }
+
+- ``getConfigurationBuilder()`` — call ``parent::getConfigurationBuilder()`` to keep the built-in
+  features (datasource, tags, categories, …), then override only what you need.
+- ``addInternalFilters()`` — the ``parent::`` call is **required**; it applies the built-in
+  ``dataSource`` and website tag/category filters. Add custom ``WHERE`` conditions after it.
+- ``addInternalSortBys()`` — called from ``addInternalFilters()`` so it shares the same
```
Prokyonn: does also need a parent:: call and is called from the parent

cookbook/custom-page-data-provider.rst
```diff
@@ -1,15 +1,102 @@
...
+
+    declare(strict_types=1);
+
+    namespace App\SmartContent;
+
+    use Doctrine\ORM\QueryBuilder;
+    use Sulu\Bundle\AdminBundle\SmartContent\Configuration\BuilderInterface;
+    use Sulu\Page\Infrastructure\Sulu\Content\PageSmartContentProvider;
+
+    readonly class FooterPagesSmartContentProvider extends PageSmartContentProvider
```
Prokyonn: Tbh I don't think it is a good example to show a "FooterPageProvider" as this indirectly suggests, that you should use a SmartContent for it, but this would be fatal, as the navigation twig extension / navigationrepository is optimized for this.

Also footer is on every page and could conflict with "content" smart content provider with the query parameters (page/filter/..)

Prokyonn (changes requested):

cookbook/custom-page-data-provider.rst
```diff
@@ -1,15 +1,143 @@
...
+            MetadataProviderInterface $formMetadataProvider,
+            SmartContentQueryEnhancer $smartContentQueryEnhancer,
+            ?TokenStorageInterface $tokenStorage,
+            EntityManagerInterface $entityManager,
+            array $bundles,
+            WebspaceManagerInterface $webspaceManager,
+            AccessControlQueryEnhancer $accessControlQueryEnhancer,
+            ?Security $security,
+            ?array $permissions,
+            private int $authorContactId,
```
Prokyonn, line 53: I would not inject this in the constructor, I think we could improve this even further, when we make use of the `$params` and set the author id/name in the xml as a parameter, WDYT?

PR #926, opened 2026-07-27
Document authenticator app (TOTP) two factor authentication

wachterjohannes:
Documents the authenticator app (TOTP) two factor method, backup codes, the admin side reset (toolbar action and `sulu:security:user:reset-two-factor` command) and the force configuration introduced in sulu/sulu#8959.

Credits to Claus Haselbach and the team at adesso, whose production-tested bundle served as the reference for the feature.


draconivis (approved):

wachterjohannes:
Updated the branch to also mention the `google` authenticator app method added in sulu/sulu#8972 — the section now covers both `scheb/2fa-totp` and `scheb/2fa-google-authenticator`. If sulu/sulu#8972 should not make it in, the mention can be dropped again.

draconivis:
LGTM

bundles/security/index.rst
```diff
@@ -220,6 +235,50 @@ the ``config/routes/scheb_2fa.yaml`` file:
...
+
+Administrators can reset the two-factor authentication of a user, for example when the user lost
+access to the authenticator app. This is possible via the ``Reset Two-Factor authentication``
+toolbar button on the permissions tab of the user or via the following command:
+
+.. code-block:: bash
+
+    bin/adminconsole sulu:security:user:reset-two-factor <username>
+
+The user can then log in with the password only and set up two-factor authentication again.
```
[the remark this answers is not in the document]
wachterjohannes: Good point — documented in 4d0dc90: the reset section now states that for users matched by forced two-factor authentication the email method is activated again automatically, so the reset can not be used to bypass the enforced second factor.

bundles/security/index.rst
```diff
@@ -220,6 +242,52 @@ the ``config/routes/scheb_2fa.yaml`` file:
...
+access to the authenticator app. This is possible via the ``Reset Two-Factor authentication``
+toolbar button on the permissions tab of the user or via the following command:
+
+.. code-block:: bash
+
+    bin/adminconsole sulu:security:user:reset-two-factor <username>
+
+The user can then log in with the password only and set up two-factor authentication again. If
+forced two-factor authentication (see below) applies to the user, the email method is activated
+again automatically instead, so the reset can not be used to bypass the enforced second factor.
```
Prokyonn: The last two sentences don't match the current behaviour - the reset **does** clear an enforced second factor.

Both reset paths (`ResetTwoFactorCommand` and `UserManager::resetTwoFactor()`) only do `setTwoFactor(null)` + `remove($twoFactor)` + `flush()`. The only place that re-activates `email` is `ForceTwoFactorSubscriber`, which listens on Doctrine `prePersist`/`preUpdate` of `User`. Since `User::$twoFactor` is the inverse side of the association (`mapped-by="user"`, FK `idUsers` on `se_user_two_factors`), nulling it never dirties the `User` row, so `preUpdate` doesn't fire and the subscriber never runs.

Reproduced with `force.enabled: true` and `pattern: '/@example\.com$/'`: a newly created `forceprobe@example.com` got `method=email` automatically, and after `bin/adminconsole sulu:security:user:reset-two-factor forceprobe` the `se_user_two_factors` table was empty again - nothing was re-created, also not on the next `User` update.

```suggestion
The user can then log in with the password only and set up two-factor authentication again.
```

If the documented behaviour is the intended one, this should be fixed in sulu/sulu instead - happy to open an issue for it.

bundles/security/index.rst
```diff
@@ -220,6 +242,52 @@ the ``config/routes/scheb_2fa.yaml`` file:
...
+.. code-block:: yaml
+
+    sulu_security:
+        two_factor:
+            force:
+                enabled: true
+                pattern: '/@sulu\.io$/'
+
+The email method is activated automatically for matching users that have no two-factor method
+configured yet. A user that has already configured a two-factor method is not affected.
```
Prokyonn: This only holds for newly created users. The subscriber persists the `UserTwoFactor` inside the Doctrine `preUpdate` callback, but Doctrine runs `executeInserts()` before `executeUpdates()` in a commit, so that insert is discarded - I flushed a `User` update with force enabled and no existing two-factor row (what `LastLoginListener` does on every interactive login) and no row was written, also not on a following flush. Only the `prePersist` path works.

So enabling the configuration in an existing project does not retrofit existing users.

```suggestion
The email method is activated automatically when a matching user is created and has no
two-factor method configured yet. A user that has already configured a two-factor method is
not affected. Users that already exist when the configuration is enabled are not migrated.
```

Prokyonn: Worth documenting the part that actually enforces the second factor: `ProfileTwoFactorController::deleteAction()` rejects the request with an `AccessDeniedHttpException` when the user's email matches the pattern, so an affected user can not switch two-factor authentication off in their own profile.

Suggestion as a plain block, because it overlaps with the suggestion above - to be appended after this paragraph:

```rst
A user that is affected by the forced configuration can not disable two-factor authentication
in the profile, the server rejects the request.
```

Prokyonn (commented):
Checked the documented behaviour against the code of the 2.6 branch including sulu/sulu#8972. Two things do not match, plus one optional addition.

bundles/security/index.rst
```diff
@@ -220,6 +242,54 @@ the ``config/routes/scheb_2fa.yaml`` file:
...
+            force:
+                enabled: true
+                pattern: '/@sulu\.io$/'
+
+The email method is activated automatically when a matching user is created and has no
+two-factor method configured yet. A user that has already configured a two-factor method is
+not affected. Users that already exist when the configuration is enabled are not migrated.
+
+A user that is affected by the forced configuration can not disable two-factor authentication
+in the profile, the server rejects the request.
```
Prokyonn: My comment above was wrong, sorry - this paragraph should go again.

The `AccessDeniedHttpException` is in `ProfileTwoFactorController::deleteAction()` (`DELETE /admin/api/profile/two-factor`), but the administration interface never calls that endpoint - `SecurityAdmin::getConfig()` only exposes `twoFactorSetup`, `twoFactorConfirm` and `twoFactorBackupCodes`. Selecting ``None`` in the profile only changes the form value, which is saved through `PUT /admin/api/profile`, and `ProfileController::putAction()` has no force check: its else branch removes the `UserTwoFactor` unconditionally.

Verified with a user matched by the force pattern:

* `PUT /admin/api/profile` with `twoFactor.method = null` -> `200`, the row is gone from `se_user_two_factors`
* `DELETE /admin/api/profile/two-factor` -> `403`

```suggestion
```

Prokyonn (commented):
Correction to one of my comments below.

Prokyonn (approved):

PR #934, opened 2026-08-22
Document the parallel image generation limit

Amoifr:
| Q | A
| --- | ---
| Fixed tickets | -
| Related PRs | sulu/sulu#9041 |
| License | MIT

#### What's in this PR?

Documents the `sulu_media.format_manager.max_parallel_image_generation_limit` and `max_parallel_image_generation_wait_time` options added in sulu/sulu#9041 (sulu/sulu#8735): a "Limiting the parallel image generation" section on the media configuration page, with the problem it solves, the YAML snippet, the Semaphore/Lock requirements and the wait behavior. To be merged together with (or after) the feature PR.

#### Why?

New configuration options, otherwise undiscoverable.

### DIFF
--- a/bundles/media/configuration.rst
+++ b/bundles/media/configuration.rst
@@ -40,17 +40,25 @@ free slot before their image is generated:
 
 The limit is shared between all the PHP workers of the application through a
 semaphore of the `Symfony Semaphore component`_, using the storage configured
-under ``framework.semaphore`` (e.g. a Redis DSN):
+under ``framework.semaphore``. The ``lock://`` storage works out of the box,
+without any extra service:
 
 .. code-block:: bash
 
-    composer require symfony/semaphore
+    composer require symfony/semaphore symfony/lock
 
 .. code-block:: yaml
 
+    # config/packages/lock.yaml
+    framework:
+        lock: '%env(LOCK_DSN)%'
+
     # config/packages/semaphore.yaml
     framework:
-        semaphore: 'redis://localhost'
+        semaphore: 'lock://'
+
+The ``lock://`` storage requires ``symfony/semaphore`` 8.1. On older versions,
+configure a Redis DSN instead (``semaphore: 'redis://localhost'``).
 
 When no slot becomes available within one minute, the request fails: the limit
 protects the server, generating the image anyway would defeat it.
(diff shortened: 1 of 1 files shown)


bundles/media/configuration.rst
```diff
@@ -18,3 +18,41 @@ The SuluMediaBundle can be configured the following way:
...
+
+.. code-block:: bash
+
+    composer require symfony/semaphore
+
+.. code-block:: yaml
+
+    # config/packages/semaphore.yaml
+    framework:
+        semaphore: 'redis://localhost'
```
alexander-schranz: I would here go with the `lock://` as that is the one which works out of the box without additional redis:

```suggestion
.. code-block:: bash

    composer require symfony/semaphore symfony/lock

.. code-block:: yaml

    # config/packages/lock.yaml
    framework:
        lock: '%env(LOCK_DSN)%'

    # config/packages/semaphore.yaml
    framework:
        semaphore: 'lock://'
```

But I'm open to keep it also as it is. 

Currently it seems symfony does not provide any recipe for semaphore, so unsure if Semaphore should be a DSN like the Lock or not. But that something which need be decide on Symfony Recipe side.

Amoifr: Applied, with one addition: `lock://` maps to the `LockStore` that needs `symfony/semaphore` 8.1 (symfony/symfony#59202; FrameworkBundle resolves `lock://<name>` to the named lock factory since then). Older Symfony versions only support Redis DSNs for the semaphore, and sulu supports 6.4+, so I kept a line pointing them to `redis://`. Funny enough, that store is exactly what the first version of the code PR used. 😄

bundles/media/configuration.rst
```diff
@@ -18,3 +18,51 @@ The SuluMediaBundle can be configured the following way:
                 - video/quicktime
                 - video/x-msvideo
                 - video/x-ms-wmv
+
+Limiting the parallel image generation
+--------------------------------------
+
+Image formats are generated on the fly the first time they are requested. When
```
Prokyonn: as we only have the 3.x branch for all 3.0/3.1/... documentation, we have to add here that this feature is only available from 3.1

Prokyonn (changes requested):

bundles/media/configuration.rst
Amoifr: Good point, thanks! Added "Available since Sulu 3.1." at the top of the section (ce6df97), same wording as the snippet template groups.

Prokyonn (approved):

Prokyonn:
@Amoifr Thank you :)

PR #937, opened 2026-09-08
Document template groups for articles and snippets

wachterjohannes:
Adds a Template Groups section to book/templates.rst covering the <group> element (article and snippet templates only, not pages): how it splits the list into per-group tabs, the per-group security contexts, and filtering selection fields/smart content by group. Adds a short pointer with an example in bundles/snippet.rst.

Covers both the article-side feature (sulu/sulu#8124, previously undocumented) and the snippet-side feature (sulu/sulu#9065).

### DIFF
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -949,6 +949,74 @@ node and the ``ref`` attribute:
     to avoid confusion. This approach also simplifies the transition to global blocks in the future, eliminating the need
     for data migrations.
 
+.. _templates-template-groups:
+
+Template Groups
+----------------
+
+.. note::
+
+    Template groups are available for article and snippet templates only. Page templates do not
+    support the ``<group>`` element.
+
+If you have many templates of the same type, you can organize them into groups by adding a
+``<group>`` element next to the ``<key>``:
+
+.. code-block:: xml
+
+    <!-- config/templates/articles/news.xml -->
+    <?xml version="1.0" ?>
+    <template xmlns="http://schemas.sulu.io/template/template"
+              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">
+
+        <key>news</key>
+        <group>news</group>
+
+        <!-- ... -->
+    </template>
+
+Templates without a ``<group>`` element land in the implicit ``default`` group.
+
+Grouping templates this way changes the administration interface for the corresponding resource
+(articles or snippets):
+
+* The list is split into one tab per group, each showing only the entries that use a template from
+  that group.
+* The "Add" and "Edit" forms of a group only offer the templates belonging to that group.
+* Every group other than ``default`` registers its own security context, e.g.
+  ``sulu.article.articles_news`` or ``sulu.snippet.snippets_news``, so access can be granted or
+  denied per group. See :doc:`../bundles/security/index` for more on security contexts. Remember to
+  grant the new context to the roles that need it. The ``default`` group and setups with only one
+  group keep using the resource's regular security context (``sulu.article.articles`` /
+  ``sulu.snippet.snippets``).
+* Selection fields such as |snippet_selection|, |single_snippet_selection|,
+  :doc:`article_selection <../reference/property-types/article_selection>` or
+  :doc:`single_article_selection <../reference/property-types/single_article_selection>`, and the
+  article/snippet smart content, can be restricted to specific groups with the ``groups``
+  parameter:
+
+.. code-block:: xml
+
+    <property name="relatedNews" type="article_selection">
+        <meta>
+            <title lang="en">Related News</title>
+        </meta>
+        <params>
+            <param name="groups" value="news"/>
+        </params>
+    </property>
+
+The group title shown in the administration interface comes from the translation key
+``sulu_admin.template_group.<group>`` and falls back to the capitalized group identifier if no
+translation exists.
+
+.. note::
+
+    Groups are also picked up by the admin search index. After adding or renaming a group, run
+    ``bin/console cmsig:seal:reindex`` for both kernels so search results deep-link into the correct
+    group.
+
 Caching
 -------
 
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -981,8 +981,8 @@ context (e.g. ``sulu.article.articles_marketing``, see :doc:`../bundles/security
 be granted to the roles that need it; selection fields and smart content for the resource can be
 filtered by group too, see e.g. :doc:`../reference/property-types/snippet_selection`.
 
-The group title shown in the administration interface — for every group, including the implicit
-``default`` group — comes from the translation key ``sulu_admin.template_group.<group>`` and falls
+The group title shown in the administration interface, for every group including the implicit
+``default`` group, comes from the translation key ``sulu_admin.template_group.<group>`` and falls
 back to the capitalized group identifier if no translation exists.
 
 .. note::
--- a/book/templates.rst
+++ b/book/templates.rst
@@ -973,9 +973,7 @@ the same type, you can organize them into groups by adding a ``<group>`` element
         <!-- ... -->
     </template>
 
-Templates without a ``<group>`` element land in the implicit ``default`` group, whose tab title
-falls back to the plain word "Default" unless your project adds a translation for
-``sulu_admin.template_group.default`` (see below).
+Templates without a ``<group>`` element land in the implicit ``default`` group.
 
 Grouping templates this way splits the list into one tab per group, restricts each group's "Add" and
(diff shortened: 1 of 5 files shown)


book/templates.rst
```diff
@@ -949,6 +949,74 @@ node and the ``ref`` attribute:
     to avoid confusion. This approach also simplifies the transition to global blocks in the future, eliminating the need
     for data migrations.
 
+.. _templates-template-groups:
+
+Template Groups
+----------------
+
+.. note::
```
Prokyonn: I dont think it makes sense to start with a note directly after the headline, I would just add it as normal text

book/templates.rst
```diff
@@ -949,6 +949,74 @@ node and the ``ref`` attribute:
...
+.. code-block:: xml
+
+    <!-- config/templates/articles/news.xml -->
+    <?xml version="1.0" ?>
+    <template xmlns="http://schemas.sulu.io/template/template"
+              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">
+
+        <key>news</key>
+        <group>news</group>
```
Prokyonn: use another name that is not the same as the key for the example, otherwise people may think they have to be connected

Prokyonn (changes requested):

book/templates.rst
```diff
@@ -949,6 +949,74 @@ node and the ``ref`` attribute:
...
     for data migrations.
 
+.. _templates-template-groups:
+
+Template Groups
+----------------
+
+.. note::
+
+    Template groups are available for article and snippet templates only. Page templates do not
```
Prokyonn: mention that snippet support that only in >=3.1

bundles/snippet.rst
```diff
@@ -102,6 +102,24 @@ Properties
...
+
+.. code-block:: xml
+
+    <!-- config/templates/snippets/social_media.xml -->
+    <key>social_media</key>
+    <group>marketing</group>
+
+See :ref:`templates-template-groups` for the full explanation, including the ``groups`` parameter for
+:doc:`../reference/property-types/snippet_selection` and
+:doc:`../reference/property-types/single_snippet_selection`, and how groups affect security contexts.
```
wachterjohannes, line 120: add to the (single_)snippet_selection that it can be filtered by group

book/templates.rst
```diff
@@ -949,6 +949,74 @@ node and the ``ref`` attribute:
...
+* Every group other than ``default`` registers its own security context, e.g.
+  ``sulu.article.articles_news`` or ``sulu.snippet.snippets_news``, so access can be granted or
+  denied per group. See :doc:`../bundles/security/index` for more on security contexts. Remember to
+  grant the new context to the roles that need it. The ``default`` group and setups with only one
+  group keep using the resource's regular security context (``sulu.article.articles`` /
+  ``sulu.snippet.snippets``).
+* Selection fields such as |snippet_selection|, |single_snippet_selection|,
+  :doc:`article_selection <../reference/property-types/article_selection>` or
+  :doc:`single_article_selection <../reference/property-types/single_article_selection>`, and the
+  article/snippet smart content, can be restricted to specific groups with the ``groups``
```
Prokyonn: trim this full block to one sentence and add the group filter for 3.1 as a note in the snippet selections

book/templates.rst
```diff
@@ -949,6 +949,74 @@ node and the ``ref`` attribute:
...
+
+        <key>news</key>
+        <group>news</group>
+
+        <!-- ... -->
+    </template>
+
+Templates without a ``<group>`` element land in the implicit ``default`` group.
+
+Grouping templates this way changes the administration interface for the corresponding resource
```
wachterjohannes: colapse the rest into a sentence and move the group filter stuff to the selection reference doc

book/templates.rst
wachterjohannes: Fixed in a35abc7: turned it into plain text at the start of the paragraph instead.

book/templates.rst
wachterjohannes: Fixed in a35abc7: example now uses key `event` with group `marketing`.

book/templates.rst
wachterjohannes: Added in a35abc7: articles since Sulu 3.0, snippets since Sulu 3.1.

book/templates.rst
wachterjohannes: Fixed in a35abc7: trimmed the whole effects list to one sentence and moved the selection-field detail to snippet_selection.rst/single_snippet_selection.rst as new `groups` parameter rows, noting the 3.1 requirement there.

book/templates.rst
wachterjohannes: Fixed in a35abc7, same as the reply above: collapsed into one sentence, selection-field detail moved to the property type reference pages.

bundles/snippet.rst
wachterjohannes: Added in a35abc7: both snippet_selection.rst and single_snippet_selection.rst get a new `groups` parameter row (available since Sulu 3.1). While there I also noticed the documented `types` parameter is stale, it was replaced by `templateKeys` a while ago and now actually fails the container build, so I fixed that too.

book/templates.rst
```diff
@@ -949,6 +949,74 @@ node and the ``ref`` attribute:
...
+  ``sulu.article.articles_news`` or ``sulu.snippet.snippets_news``, so access can be granted or
+  denied per group. See :doc:`../bundles/security/index` for more on security contexts. Remember to
+  grant the new context to the roles that need it. The ``default`` group and setups with only one
+  group keep using the resource's regular security context (``sulu.article.articles`` /
+  ``sulu.snippet.snippets``).
+* Selection fields such as |snippet_selection|, |single_snippet_selection|,
+  :doc:`article_selection <../reference/property-types/article_selection>` or
+  :doc:`single_article_selection <../reference/property-types/single_article_selection>`, and the
+  article/snippet smart content, can be restricted to specific groups with the ``groups``
+  parameter:
```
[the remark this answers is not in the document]
wachterjohannes: Good catch, that reference page predates the snippet-side feature. Fixed in a35abc7: smart_content.rst's `groups` row now lists the `snippets` provider too (since Sulu 3.1), verified against SnippetSmartContentProvider::resolveTemplateKeys() which intersects `groups` with `templateKeys` exactly like the article provider does.

book/templates.rst
```diff
@@ -949,6 +949,48 @@ node and the ``ref`` attribute:
...
+              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">
+
+        <key>event</key>
+        <group>marketing</group>
+
+        <!-- ... -->
+    </template>
+
+Templates without a ``<group>`` element land in the implicit ``default`` group.
```
Prokyonn, line 976: add a note about the translation key

wachterjohannes: Added in 8ed2490: noted that the default group follows the same translation key convention (`sulu_admin.template_group.default`), and since Sulu does not ship a translation for it, it shows as the plain word "Default" unless the project adds one. Let me know if you meant something else by this.

book/templates.rst
```diff
@@ -973,7 +973,9 @@ the same type, you can organize them into groups by adding a ``<group>`` element
         <!-- ... -->
     </template>
 
-Templates without a ``<group>`` element land in the implicit ``default`` group.
+Templates without a ``<group>`` element land in the implicit ``default`` group, whose tab title
+falls back to the plain word "Default" unless your project adds a translation for
+``sulu_admin.template_group.default`` (see below).
```
Prokyonn: this applies for every group

book/templates.rst
wachterjohannes: You are right, that was misleading. Fixed in 0a30da5: dropped the default-specific sentence and instead made the existing general statement explicit that it applies to every group, including default, not just custom ones.

book/templates.rst
```diff
@@ -973,19 +973,17 @@ the same type, you can organize them into groups by adding a ``<group>`` element
...
 Grouping templates this way splits the list into one tab per group, restricts each group's "Add" and
 "Edit" forms to that group's templates, and gives every group other than ``default`` its own security
 context (e.g. ``sulu.article.articles_marketing``, see :doc:`../bundles/security/index`) that has to
 be granted to the roles that need it; selection fields and smart content for the resource can be
 filtered by group too, see e.g. :doc:`../reference/property-types/snippet_selection`.
 
-The group title shown in the administration interface comes from the translation key
-``sulu_admin.template_group.<group>`` and falls back to the capitalized group identifier if no
-translation exists.
+The group title shown in the administration interface — for every group, including the implicit
```
Prokyonn: remove the dashes and the ;

Prokyonn (approved):

